PR #1, opened 2016-04-08
[JENKINS-26156] Do not call BodyInvoker.withDisplayName(null)

jglick:
See https://github.com/jenkinsci/workflow-cps-plugin/pull/3 for discussion.

@reviewbybees


amuniz:
:bee:

PR #2, opened 2016-04-09
[JENKINS-26132] Show a label in the executor cell

jglick:
[JENKINS-26132](https://issues.jenkins-ci.org/browse/JENKINS-26132)

Adjusts UI of the queue and executor widgets to show label information. For a `node` block waiting in queue, this would show the nearest label enclosing the block, if any; for one executing, this would show the nearest label enclosing the currently running step, if any. Labels could be `stage` names or `parallel` branch names.

Interactively tested in `workflow-aggregator-plugin/demo/` and looks reasonable. Here is one branch of the testing stage running and one waiting for a second agent to come online:

widgets

Here are a bunch of builds running in various stages (note that the **master** flyweight executor represents running Groovy code):

widgets2

@reviewbybees


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly
```diff
@@ -24,14 +24,18 @@
...
         <div style="white-space: normal">
             <j:set var="r" value="${it.parent.runForDisplay()}"/>
             <j:set var="rAuth" value="${r != null and h.hasPermission(r.parent, r.parent.READ) ? r : null}"/>
             <j:choose>
                 <j:when test="${rAuth != null}">
-                    part of <a href="${rootURL}/${r.parent.url}"><l:breakable value="${r.parent.fullDisplayName}"/></a>
+                    <a href="${rootURL}/${r.parent.url}"><l:breakable value="${r.parent.fullDisplayName}"/></a>
+                    <j:set var="label" value="${it.parent.enclosingLabel}"/>
+                    <j:if test="${label != null}">
+                        <st:nbsp/>(${label})
```
amuniz: IIUC, this example:

```groovy
stage 'compile'
load 'external.groovy' // external groovy contains: node { echo 'hi' }
```
will show `external.groovy`, and this:

```
stage 'compile'
node {
    echo 'hi'
}
```

will show `compile`, right?

amuniz: IMO from a user point of view, both should show `compile`.

jglick: > …right?

Right. More generally, it will show the most specific label applicable to the running step.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -329,6 +331,35 @@ private Object readResolve() {
...
+        public @CheckForNull String getEnclosingLabel() throws Exception {
+            FlowNode executorStepNode = context.get(FlowNode.class);
+            if (executorStepNode != null) {
+                for (StepExecution exec : executorStepNode.getExecution().getCurrentExecutions(true).get(1, TimeUnit.MINUTES)) {
+                    FlowNode runningNode = exec.getContext().get(FlowNode.class);
+                    if (runningNode != null) {
+                        // See if this step is inside our node {} block, and track the associated label.
+                        boolean match = false;
+                        String label = null;
+                        FlowNodeSerialWalker.EnhancedIterator it = new FlowNodeSerialWalker(runningNode).iterator();
```
svanoort: @jglick  Point of consideration: there may be many steps ahead of this before we see our first labeled or stage node.  Is there some form of implicit caching here that I am missing?  Otherwise, we need to make an effort not to walk the same flow graph with every UI view. 

My suggestion: store the ID & label of the last node in we got the enclosing label for.  If it is unchanged, return it.  Otherwise, only iterate through the walker until we hit changes in the node, or we hit the last node again.

Not perfect because of parallel blocks that can contain node blocks and labels, but better at least.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -329,6 +331,35 @@ private Object readResolve() {
...
+                        while (it.hasNext()) {
+                            FlowNode n = it.next();
+                            if (label == null) {
+                                label = it.currentLabel();
```
svanoort: :ant:  IMO easier to follow if you make an explicit check for the iterator's current label not being null before assignment, and add a comment explaining.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -329,6 +331,35 @@ private Object readResolve() {
...
+                            }
+                            if (n.equals(executorStepNode)) {
+                                match = true;
```
svanoort: :bug:  break on match and label != null.  Otherwise we walk the entire flow graph, no?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
svanoort: :bug:  I believe we should break out once we have both a label and and a match, otherwise we walk all the way to the beginning of the flow, no?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -329,6 +331,35 @@ private Object readResolve() {
             return getDisplayName();
         }
 
+        @Restricted(NoExternalUse.class) // for Jelly
+        public @CheckForNull String getEnclosingLabel() throws Exception {
+            FlowNode executorStepNode = context.get(FlowNode.class);
+            if (executorStepNode != null) {
+                for (StepExecution exec : executorStepNode.getExecution().getCurrentExecutions(true).get(1, TimeUnit.MINUTES)) {
```
svanoort: This feels really iffy to me, but I can't put a finger on why.   But I do wonder if there is a reason why we would want to wait a minute while doing UI rendering?

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly
```diff
@@ -24,14 +24,18 @@
...
     <td class="pane">
         <div style="white-space: normal">
             <j:set var="r" value="${it.parent.runForDisplay()}"/>
             <j:set var="rAuth" value="${r != null and h.hasPermission(r.parent, r.parent.READ) ? r : null}"/>
             <j:choose>
                 <j:when test="${rAuth != null}">
-                    part of <a href="${rootURL}/${r.parent.url}"><l:breakable value="${r.parent.fullDisplayName}"/></a>
+                    <a href="${rootURL}/${r.parent.url}"><l:breakable value="${r.parent.fullDisplayName}"/></a>
+                    <j:set var="label" value="${it.parent.enclosingLabel}"/>
+                    <j:if test="${label != null}">
```
svanoort: Thoughts here: what will happen to the layout if there is a very long enclosing label?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -422,4 +424,36 @@ private void startJnlpProc() throws Exception {
         });
     }
 
+    @Issue("JENKINS-26132")
+    @Test public void enclosingLabel() {
```
svanoort: Looks pretty solid to me;  in the future we may wish to add coverage of nested  & parallel labels just to ensure that works correctly, but for stages... not an issue.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: We expect `get` to return immediately; the timeout is just to prevent a UI thread from hanging if something weird goes wrong.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: > we need to make an effort not to walk the same flow graph with every UI view

Why bother? This is a running build, and the whole graph should be in memory. Would be worth capping the number of nodes traversed at, say, 100, in case of a big graph.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Yeah, could do that optimization.

src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly
jglick: Should probably use `l:breakable` on the label.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
jglick: Until https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/1 is merged, it is hard to test nested labels. Could so with `parallel`, I suppose. But the heavier tests of that aspect are already in the upstream PR.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
svanoort: If we expect immediate return in all cases but errors,  wouldn't it make sense to reduce the timeout duration to something fairly short (like under 1s)?  Many users are likely to see this UI element rendered at once, so 1 min is long enough that we could still exhaust thread pools. 

I'm assuming that in the event of the timeout we're fine with just letting the TimeoutException bubble up, if it's a rare event, so not as concerned about that part.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
svanoort: > Why bother?

Normally I'd be all for keeping it simple but we've already gotten bitten by some of these assumptions.  In this case I'd ask  why force the master to do extra work with every UI paint, when it's trivial to avoid it?  

Also, what forces the entire flowgraph to stay in memory, from what I am seeing, shouldn't it just be head nodes + whatever is in the soft-reference cache of flownodes in the xstream FlowNodeStorage, or am I missing a set of full hard references?

jglick: > am I missing a set of full hard references?

At least for builds initiated during the current Jenkins session, the whole graph is in memory, via `FlowNode.parents`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: True, could probably pick a timeout in the seconds range. IIRC the widget is refreshed every 5s, so something less than that.

jglick:
Since we have no definite conclusion yet from https://github.com/jenkinsci/workflow-api-plugin/pull/1 I think I will just copy `FlowNodeSerialWalker` here, make it `@Restricted(NoExternalUse.class)`. and consider this ready for final review—can switch to any new graph walking API when it appears.

svanoort:
@jglick If you like, but one does need a *little* time to work. ;)  I think I have the code here done, just need to do tests and add some convenience methods for this.  Maybe clean it up a bit.

jglick:
Needs to be updated to use the new APIs.

jglick:
@reviewbybees done

CC @svanoort who commented on its original incarnation

svanoort:
Yeah, I'll assign myself as a reviewer and want to take a closer look before we cut this one

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -405,7 +411,17 @@ private Object readResolve() {
         @Override public String getDisplayName() {
             // TODO more generic to check whether FlowExecution.owner.executable is a ModelObject
             Run<?,?> r = runForDisplay();
-            return r != null ? Messages.ExecutorStepExecution_PlaceholderTask_displayName(r.getFullDisplayName()) : Messages.ExecutorStepExecution_PlaceholderTask_displayName_unknown();
+            if (r != null) {
+                String runDisplayName = r.getFullDisplayName();
+                String label = getEnclosingLabel();
+                if (label != null) {
```
oleg-nenashev: Could it be empty/blank? AFAIK yes for the label strings in the core

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -416,6 +432,70 @@ private Object readResolve() {
...
+                        }
+                    }
+                    if (n.equals(executorStepNode)) {
+                        if (label != null) {
+                            return label;
+                        }
+                        match = true;
+                    }
+                    if (count++ > 100) {
+                        break; // not important enough to bother
```
oleg-nenashev: Hmm, really?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -416,6 +432,70 @@ private Object readResolve() {
             return getDisplayName();
         }
 
+        /** hash code of list of heads */
+        private transient int lastCheckedHashCode;
+        private transient String lastEnclosingLabel;
+        @Restricted(NoExternalUse.class) // for Jelly
+        public @CheckForNull String getEnclosingLabel() {
```
oleg-nenashev: I would synchronize this method or offload it to the async operation (via AJAX). It includes an operation with a 100 ms timeout, hence the current implementation may consume all Jenkins executors if there are many ExecutorStep rendering requests

jglick: In fact I expect the implementation to be instantaneous. There was a bug in an earlier iteration which I corrected here and also in https://github.com/jenkinsci/workflow-cps-plugin/issues/145.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: No a label should not be blank.

You are perhaps confusing this with agent labels. Totally unrelated.

jglick:
> want to take a closer look

Does not seem that this is going to happen any time soon, so proceeding without. Anyway this is just UI tweaks, not APIs of any kind, so it can be freely amended or even reverted later if there are any concerns.

jglick:
Had some minor additional changes, so put back up for review.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -405,7 +411,17 @@ private Object readResolve() {
         @Override public String getDisplayName() {
             // TODO more generic to check whether FlowExecution.owner.executable is a ModelObject
             Run<?,?> r = runForDisplay();
-            return r != null ? Messages.ExecutorStepExecution_PlaceholderTask_displayName(r.getFullDisplayName()) : Messages.ExecutorStepExecution_PlaceholderTask_displayName_unknown();
+            if (r != null) {
+                String runDisplayName = r.getFullDisplayName();
+                String enclosingLabel = getEnclosingLabel();
```
jglick: Unfortunately as per [JENKINS-45553](https://issues.jenkins-ci.org/browse/JENKINS-45553), this is also called by `getFullDisplayName`, which `Queue.maintain` calls frequently, and there is not yet an alternate API to be used for this purpose. Options:

* override `getFullDisplayName` to omit the label, and hope there are no important UI uses (`executorCell` does its own thing, but would need to check the queue widget)
* define the core API and use that here (adding `@Override` only when baseline bumped)
* somehow determine dynamically whether we are holding the `Queue.lock` (guess this would need reflection to obtain that field, followed by `isHeldByCurrentThread`)

CC @svanoort

jglick: cf. [JENKINS-36547](https://issues.jenkins-ci.org/browse/JENKINS-36547)

PR #3, opened 2016-05-02
[JENKINS-34542] Avoid holding the runningTasks lock while running Launcher.kill

jglick:
[JENKINS-34542](https://issues.jenkins-ci.org/browse/JENKINS-34542)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -351,12 +351,12 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             return cookie != null; // in which case this is after a restart and we still claim the executor
         }
 
-        private static void finish(@CheckForNull String cookie) {
+        private static void finish(@CheckForNull final String cookie) {
             if (cookie == null) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -370,17 +370,17 @@ private static void finish(@CheckForNull String cookie) {
                 Timer.get().submit(new Runnable() { // JENKINS-31614
                     @Override public void run() {
                         execution.completed(null);
+                        try {
+                            runningTask.launcher.kill(Collections.singletonMap(COOKIE_VAR, cookie));
+                        } catch (ChannelClosedException x) {
+                            // fine, Jenkins was shutting down
+                        } catch (RequestAbortedException x) {
+                            // slave was exiting; too late to kill subprocesses
```
amuniz: Log something here?

amuniz:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Could perhaps log something at `FINE`; it is not considered a problem.

BTW this code is not new, merely reindented.

PR #4, opened 2016-06-02
[JENKINS-34281] Better robustness against missing queue items

jglick:
[JENKINS-34281](https://issues.jenkins-ci.org/browse/JENKINS-34281) can mean that if a Pipeline build was running a `node` step but was waiting for an executor (not yet inside the block), after a restart the `Queue.Item` for the block goes missing. Without intervention, the build will just sit there quietly until interrupted.

- [X] work around JENKINS-34281
- [x] immediately fail the step if a queue item is deliberately canceled

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
@@ -37,6 +37,7 @@ import hudson.model.Executor;
 import hudson.model.Label;
 import hudson.model.Node;
 import hudson.util.FormValidation;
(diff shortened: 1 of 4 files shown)


jglick:
While I could perhaps write a test to cover the JENKINS-34281 workaround, (a) it would not do anything meaningful as of `-Djenkins.version=2.1`, (b) I am not sure how to simulate an “impolite” restart from `RestartableJenkinsRule`: `JenkinsRule.after` does call `Jenkins.cleanUp`, which saves the queue.

amuniz:
:bee:

andresrc:
:bee:

PR #6, opened 2016-06-09
[JENKINS-26130] Way to print progress from pickles

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/5, https://github.com/jenkinsci/workflow-support-plugin/pull/4, https://github.com/jenkinsci/workflow-job-plugin/pull/5, and https://github.com/jenkinsci/workflow-cps-plugin/pull/20.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -104,15 +104,15 @@ public class ExecutorPickle extends Pickle {
             @Override protected void printWaitingMessage(TaskListener listener) {
                 Queue.Item item = Queue.getInstance().getItem(itemID);
                 if (item == null) { // ???
-                    listener.getLogger().println("Waiting to start " + task.getFullDisplayName());
+                    listener.getLogger().println("Waiting to resume " + task.getFullDisplayName());
                     return;
                 }
                 CauseOfBlockage causeOfBlockage = item.getCauseOfBlockage();
                 if (causeOfBlockage != null) {
-                    listener.getLogger().print("Waiting to start " + task.getFullDisplayName() + ": ");
+                    listener.getLogger().print("Waiting to resume " + task.getFullDisplayName() + ": ");
                     causeOfBlockage.print(listener); // note that in case of Messages.Queue_Unknown for WaitingItem this is not very helpful
                 } else {
-                    listener.getLogger().println("Waiting to start " + task.getFullDisplayName());
+                    listener.getLogger().println("Waiting to resume " + task.getFullDisplayName());
                 }
             }
         };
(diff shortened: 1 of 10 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -60,21 +67,29 @@ private ExecutorPickle(Executor executor) {
...
                     if (item == null) {
                         // TODO should also report when !ScheduleResult.created, since that is arguably an error
                         throw new IllegalStateException("queue refused " + task);
                     }
-                    future = item.getFuture().getStartCondition();
+                    itemID = item.getId();
+                } else {
+                    item = Queue.getInstance().getItem(itemID);
+                    if (item == null) {
+                        throw new IllegalStateException("queue lost item #" + itemID);
```
oleg-nenashev: I'm not sure if it's safe if the QueueTask gets completed quickly

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -88,6 +103,33 @@ protected Executor tryResolve() throws Exception {
...
+                }
+            }
+            @Override public boolean cancel(boolean mayInterruptIfRunning) {
+                Queue.Item item = Queue.getInstance().getItem(itemID);
+                if (item != null) {
+                    if (!Queue.getInstance().cancel(item)) {
+                        LOGGER.log(Level.WARNING, "failed to cancel {0}", item);
+                    }
+                }
+                return super.cancel(mayInterruptIfRunning);
```
oleg-nenashev: Is it really required?

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
jglick: `Queue.leftItems` keeps things for five minutes, much longer than the recurrence period for this pickle.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
jglick: Which part? The `Queue.cancel` call removes the _part of …_ task from the queue, which the last line of `unloadableExecutorPickle` asserts. The `super.cancel` stops the retrying in `TryRepeatedly` (though the next try would probably fail anyway, in `QueueTaskFuture.get`).

oleg-nenashev: ok

oleg-nenashev:
🐝

abayer:
To paraphrase Eddie Izzard, "I like my pull requests like I like my coffee - covered in :bee:s"

PR #7, opened 2016-06-15
[JENKINS-34021] If DurableTaskStep.Execution.listener is sometimes null

jglick:
[JENKINS-34021](https://issues.jenkins-ci.org/browse/JENKINS-34021)

Symptomatic & diagnostic fix.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -30,12 +30,15 @@ import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.TaskListener;
+import hudson.util.LogTaskListener;
 import java.io.IOException;
+import java.io.PrintStream;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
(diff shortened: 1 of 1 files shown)


oleg-nenashev:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -114,13 +118,33 @@ protected DurableTaskStepDescriptor() {
...
+            TaskListener l = listener;
+            if (l == null) {
+                StepContext context = getContext();
+                try {
+                    l = context.get(TaskListener.class);
+                    if (l != null) {
+                        LOGGER.log(Level.WARNING, "JENKINS-34021: DurableTaskStep.Execution.listener not restored in {0}", context);
+                    } else {
+                        LOGGER.log(Level.WARNING, "JENKINS-34021: TaskListener not even available upon request in {0}", context);
+                        l = new LogTaskListener(LOGGER, Level.FINE);
```
amuniz: This is redirecting the log messages to the file logger, right? Maybe worth noting it in the previous WARNING log.

amuniz:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Well, if this ever happens again, a dev is going to need to supply user instructions on enabling logging etc.

PR #10, opened 2016-07-12
A few cleanups in tests

jglick:
@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
@@ -106,9 +106,6 @@ public class ShellStepTest extends Assert {
         });
 
         j.assertBuildStatus(Result.ABORTED, j.waitForCompletion(b));
-        /* TODO fails, even though WorkflowRun calls recordCauseOfInterruption:
-        assertNotNull(b.getAction(InterruptedBuildAction.class)); // would also have a UserInterruption if authenticated
-        */
     }
 
     @Test public void launcherDecorator() throws Exception {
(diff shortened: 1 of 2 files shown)


abayer:
lgtm :bee:

kohsuke:
:bee:

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -87,12 +86,9 @@ public void abort() throws Exception {
         WorkflowRun b = foo.scheduleBuild2(0).getStartCondition().get();
 
         // at this point the file should be being touched
-        waitForCond(15000, tmp, new Predicate<File>() {
-            @Override
-            public boolean apply(File tmp) {
-                return tmp.exists();
-            }
-        });
+        while (!tmp.exists()) {
```
svanoort: Like it.

svanoort: Wait, no: this really does need a timeout.

svanoort:
🐛  for removing timeout unless there is a good reason.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -87,12 +86,9 @@ public void abort() throws Exception {
...
+            Thread.sleep(100);
+        }
 
         b.getExecutor().interrupt();
```
jglick: Note that this can flake out below, probably because the interruption does not take effect synchronously, so the file might still be modified for a few moments afterward.

jglick:
> for removing timeout unless there is a good reason

There is a good reason: it was not necessary (`JenkinsRule` applies a 3m timeout anyway), and it could cause the test to flake out unnecessarily just because of system load.

svanoort:
If there's an enclosing timeout, then it sounds fine by me: 🐝

PR #11, opened 2016-07-12
[JENKINS-26133] Return exit status or stdout from sh/bat steps

jglick:
Downstream of https://github.com/jenkinsci/durable-task-plugin/pull/27.

- [X] `returnStdout`
- [x] `returnStatus`

@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -58,19 +63,70 @@
...
+        return returnStatus;
+    }
+
+    @DataBoundSetter public void setReturnStatus(boolean returnStatus) {
+        this.returnStatus = returnStatus;
+    }
+
+    public abstract static class DurableTaskStepDescriptor extends AbstractStepDescriptorImpl {
+
+        public static final String defaultEncoding = "UTF-8";
```
rsandell: _annoying nit_ constants are usually capitalized.

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: We tend to break that rule for this idiom.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -209,8 +275,8 @@ private void check() {
                         LOGGER.log(Level.FINE, "last-minute output in {0} on {1}", new Object[] {remote, node});
                     }
                     t.set(null); // do not interrupt cleanup
-                    if (exitCode == 0) {
-                        getContext().onSuccess(exitCode);
+                    if (returnStatus || exitCode == 0) {
+                        getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(controller.getOutput(workspace, launcher), encoding) : null);
                     } else {
                         getContext().onFailure(new AbortException("script returned exit code " + exitCode));
```
amuniz: I guess someone configuring `returnStdout` expects to get the stdout even when the step fails, but in the exception message maybe...

Belated :bug: I guess

jglick: Filed #13.

PR #14, opened 2016-08-02
[JENKINS-37121] Use WorkspaceList.record, not .acquire, when dehydrating a pickle

jglick:
[JENKINS-37121](https://issues.jenkins-ci.org/browse/JENKINS-37121)

The situation described in the issue is anomalous. It is better for `tryResolve` to fail immediately if the workspace was locked by something else, rather than call `acquire` and block a `Timer` thread indefinitely.

@reviewbybees


abayer:
:bee:

svanoort:
🐝

oleg-nenashev:
Maybe it would be safe to wait till the external user releases the workspace. Or does it break the Pickle rehydration concept?

jglick:
> does it break the Pickle rehydration concept?

Yes; if we had obtained a lease on the workspace before, the assumption is that we expect to be the exclusive owner after a restart as well.

oleg-nenashev:
Uhm, ok. The behavior still smells error prone for corner cases, but it's definitely better than the original code.  🐝

PR #15, opened 2016-09-06
Allow extension of ExecutorStep

ndeloof:
docker-slaves do offer pipeline `dockerNode()` and rely on a copy/paste of ExecutorStep|Execution. Not being final would allow to  rely on upstream code and benefits fixes and improvements made here

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
@@ -59,7 +59,7 @@ import java.util.Set;
  *     }
  * </pre>
  */
-public final class ExecutorStep extends AbstractStepImpl implements Serializable {
+public class ExecutorStep extends AbstractStepImpl implements Serializable {
 
     private static final long serialVersionUID = 1L;
 


jglick:
Extending `ExecutorStep` on its own is not going to do anything.

Generally I am :-1: on subclassing to accomplish things like this. Probably needs a different design.

ndeloof:
If you look at docker-slaves pipeline support you'll notice it's 99℅ à
copy/paste of this class. What would be wrong with extending it with minor
customisation?

Le 8 sept. 2016 19:15, "Jesse Glick" <notifications@github.com> a écrit :

Extending ExecutorStep on its own is not going to do anything.

Generally I am 👎 on subclassing to accomplish things like this. Probably
needs a different design.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/15#issuecomment-245689155>,
or mute the thread

ndeloof:
Quick prototype to demonstrate usage : https://github.com/Dockins/docker-slaves-plugin/tree/extends-durable-tast

Actually this doesn't work, `ExecutorStepExecution` being inherited doesn't get `@StepContextParameter` attribues injectes, seems pipeline only do consider class attributes, not inherited ones.

jglick:
Not going to take this approach.

PR #19, opened 2016-09-20
Run part details not exported via API

scoheb:
When part of a workflow run is executing on a node, the corresponding run details are not available via the API (xml/json)

Without this patch, the run part cannot be identified by examining the API call's result.

This poses a problem for an integration such as python-jenkins where its server.get_running_builds()
queries each node to build up an array of builds. As a mininum, 'number' and 'url' are needed.

The plugin exposes this data already via Jelly in the Executor view.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -62,6 +62,7 @@ import org.jenkinsci.plugins.workflow.support.actions.WorkspaceActionImpl;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.export.Exported;
 import org.kohsuke.stapler.export.ExportedBean;
 
 public class ExecutorStepExecution extends AbstractStepExecutionImpl {
@@ -479,7 +480,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
          * Occupies {@link Executor} while workflow uses this slave.
          */
         @ExportedBean
-        private final class PlaceholderExecutable implements ContinuableExecutable {
+        public final class PlaceholderExecutable implements ContinuableExecutable {
 
             @Override public void run() {
                 final TaskListener listener;
@@ -589,6 +590,23 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 return PlaceholderTask.this;
             }
 
+            @Exported
+            public int getNumber() {
+                Run<?, ?> r = getParent().runForDisplay();
+                return r != null ? r.getNumber() : -1;
+            }
+
+            @Exported
+            public String getFullDisplayName() {
+                return getParent().getFullDisplayName();
+            }
+
+            @Exported
+            public String getDisplayName() {
+                return getParent().getDisplayName();
+            }
+
+            @Exported
             @Override public long getEstimatedDuration() {
                 return getParent().getEstimatedDuration();
             }
@@ -604,9 +622,18 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 return Executor.of(this);
             }
 
-            @Restricted(NoExternalUse.class) // for Jelly and toString
+            @Exported @Restricted(NoExternalUse.class) // for Jelly and toString
             public String getUrl() {
-                return PlaceholderTask.this.getUrl(); // we hope this has a console.jelly
+                Run<?,?> r = runForDisplay();
+                if (r == null) {
+                    return "";
+                }
+                Jenkins j = Jenkins.getInstance();
+                String base = "";
+                if (j != null) {
+                    base = Util.removeTrailingSlash(j.getRootUrl()) + "/";
+                }
+                return base + r.getUrl();
             }
 
             @Override public String toString() {
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.support.steps;
 
+import com.gargoylesoftware.htmlunit.Page;
 import hudson.model.Computer;
 import hudson.model.Node;
 import hudson.model.Queue;
@@ -74,6 +75,7 @@ import org.junit.rules.TemporaryFolder;
 import org.junit.runners.model.Statement;
 import org.jvnet.hudson.test.BuildWatcher;
 import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.MockAuthorizationStrategy;
 import org.jvnet.hudson.test.RestartableJenkinsRule;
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -589,6 +590,23 @@ private static void finish(@CheckForNull final String cookie) {
                 return PlaceholderTask.this;
             }
 
+            @Exported
+            public int getNumber() {
+                Run<?, ?> r = getParent().runForDisplay();
+                return r != null ? r.getNumber() : -1;
```
jglick: Could also declare return type to be `Integer`, in which case `null` will simply skip the entry in the output.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -604,9 +622,18 @@ private static void finish(@CheckForNull final String cookie) {
...
+                Run<?,?> r = runForDisplay();
+                if (r == null) {
+                    return "";
+                }
+                Jenkins j = Jenkins.getInstance();
+                String base = "";
+                if (j != null) {
+                    base = Util.removeTrailingSlash(j.getRootUrl()) + "/";
+                }
+                return base + r.getUrl();
```
jglick: No do not touch this method, it is used [here](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/8f7dac968c97e2a123befc2419f16e24234ca785/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution/PlaceholderTask/PlaceholderExecutable/executorCell.jelly#L45) and expects a context-relative string, not a full URL.

For purposes of the REST API you want to introduce a separate method like `getAbsoluteUrl()`; `@Exported` can take an optional `name="url"` to provide the same user-visible effect.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -400,6 +402,41 @@ private void startJnlpProc() throws Exception {
...
+    @Test public void detailsExported() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                DumbSlave s = story.j.createOnlineSlave();
+
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition(
+                        "node('" + s.getNodeName() + "') {\n" +
+                                "    sleep 10\n" +
```
jglick: Wastes time in tests. Rather use the `semaphore` step and when you want to make sure we are there, use `SemaphoreStep.waitForStart`.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -400,6 +402,41 @@ private void startJnlpProc() throws Exception {
...
+                                "}"));
+
+                WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+                Thread.sleep(100);
```
jglick: Flaky, and unnecessary when using `semaphore`.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -400,6 +402,41 @@ private void startJnlpProc() throws Exception {
...
+                JenkinsRule.WebClient wc = story.j.createWebClient();
+                Page page = wc
+                        .goTo("computer/" + s.getNodeName()
+                                        + "/api/xml?depth=2&xpath=//currentExecutable/number/text()",
```
jglick: `depth` is deprecated. Do not use it. Use `tree` to select the desired fields:

    …/computer/whatever/api/xml?tree=…somethinghere…[currentExecutable[number]]

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -400,6 +402,41 @@ private void startJnlpProc() throws Exception {
...
+                assertEquals("1", page.getWebResponse().getContentAsString());
+                page = wc.goTo("computer/" + s.getNodeName()
+                                + "/api/xml?depth=2&xpath=//currentExecutable/displayName/text()",
+                        "text/plain");
+                assertEquals("part of " + p.getDisplayName() + " #1",
+                        page.getWebResponse().getContentAsString());
+                page = wc.goTo("computer/" + s.getNodeName()
+                                + "/api/xml?depth=2&xpath=//currentExecutable/url/text()",
+                        "text/plain");
+                assertEquals(story.j.getURL().toString() + "job/" + p.getDisplayName() + "/1/",
```
jglick: `.getName()` not `.getDisplayName()` (even if they happen to be the same here).

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -400,6 +402,41 @@ private void startJnlpProc() throws Exception {
...
+                Page page = wc
+                        .goTo("computer/" + s.getNodeName()
+                                        + "/api/xml?depth=2&xpath=//currentExecutable/number/text()",
+                                "text/plain");
+
+                assertEquals("1", page.getWebResponse().getContentAsString());
+                page = wc.goTo("computer/" + s.getNodeName()
+                                + "/api/xml?depth=2&xpath=//currentExecutable/displayName/text()",
+                        "text/plain");
+                assertEquals("part of " + p.getDisplayName() + " #1",
```
jglick: Test would run more quickly if you get all three fields in one response (easy enough with `tree`).

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
scoheb: I was trying to test what python-jenkins is doing with depth though...

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
scoheb: so:

`
return r != null ? r.getNumber() : null;
`

?

jglick: Right.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
jglick: If it is using `depth` then it is buggy. Better to test using `tree`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -604,9 +622,18 @@ private static void finish(@CheckForNull final String cookie) {
                 return Executor.of(this);
             }
 
-            @Restricted(NoExternalUse.class) // for Jelly and toString
+            @Exported @Restricted(NoExternalUse.class) // for Jelly and toString
```
jglick: and REST, presumably

jglick:
@svanoort @abayer anyone feel like taking this over? Would be handy.

scoheb:
not sure this is needed anymore to be honest...feel free to close it or I can

roguishmountain:
Hi @scoheb, why was this PR abandoned? I ran into this issue recently and the functionality in this PR seems to still not be part of the plugin.

dwnusbaum:
Closing in favor of #68 which builds on this PR. Thanks for getting this started @scoheb!

PR #20, opened 2016-09-20
If an executor slot is orphaned, allow it to be manually cleared

jglick:
@cyrille-leclerc came across a case where due to what seems to be massive corruption of build records on disk for unknown reasons, perhaps combined with https://github.com/jenkinsci/workflow-job-plugin/pull/26, a build was actually marked finished (no master flyweight executor, `ABORTED` status, etc.) yet it had a `PlaceholderExecutable` still occupying an executor slot, which could not be canceled. Inspection in `/script` showed that it was in this clause. Manually calling `.completed(null)` sufficed to clear the executor slot. No idea how to reproduce this situation, but it seems appropriate to do this cleanup.

@reviewbybees


oleg-nenashev:
Looks reasonable 🐝

PR #21, opened 2016-10-04
[JEP-210] Integration tests with DurableTaskStep

jglick:
Integration tests for https://github.com/jenkinsci/workflow-job-plugin/pull/27.

[JEP-210](https://github.com/jenkinsci/jep/blob/master/jep/210/README.adoc)

- [X] basic implementation passing existing tests
- [x] specific test of a remote log stream
- [X] ~automated Unicode handling test~
- [x] ~interactive Unicode handling test on Windows~
- [x] check `BindingStepTest` with a remote logger
- [x] check `WithContainerStepTest` on an agent
- [X] ~prototype against Logstash: https://github.com/jenkinsci/logstash-plugin/pull/18~


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -347,17 +391,67 @@ private void check() {
...
+        private final ExecutionRemotable execution;
+        private final TaskListener listener;
+
+        HandlerImpl(Execution execution, FilePath workspace, TaskListener listener) {
+            this.execution = workspace.getChannel().export(ExecutionRemotable.class, execution);
+            this.listener = listener;
+        }
+
+        @Override public void output(InputStream stream) throws Exception {
+            IOUtils.copy(stream, listener.getLogger());
```
jglick: This needs to be interrupted somehow when `stop` is called.

jglick:
> Would it be possible to decouple the `ShellStepStep.remoteLogger` change to a separate PR

No, that is the crucial test demonstrating the effectiveness of _this_ PR.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (from #63)

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -237,6 +258,97 @@ public DescriptorImpl() {
...
+                @Override public AnnotatedLargeText<FlowNode> stepLog(FlowNode node, boolean complete) {
+                    return base.stepLog(node, complete);
+                }
+            };
+        }
+    }
+    private static class RemotableBuildListener implements BuildListener {
+        private static final long serialVersionUID = 1;
+        /** actual implementation */
+        private final TaskListener delegate;
```
jglick: Would be nice to have another test here too, which would produce a large volume of output, using a listener that did _not_ delegate but wrote independently to a separate file, and use `Channel.Listener.onRead`/`.onWrite` to demonstrate that there is little traffic…unless you disable watch mode, as a control.

jglick: (resolved)

PR #22, opened 2016-11-23
[JENKINS-37324] Retain script argument to the batch/shell steps so we can retrieve it upon demand

svanoort:
Implements [JENKINS-37324](https://issues.jenkins-ci.org/browse/JENKINS-37324) for shell & batch step, more or less. 

This gives us the script argument that we really need to implement smarter display of batch steps.  Note that I'm not modifying the *displayName* or functionName for the step, because IIRC there is existing logic dependent on that -- but consumers of the flow graph can do the following to get what is needed:

```java
ScriptArgumentAction action = flowNode.getAction(ScriptArgumentAction.class);
String batchOrShellScript = action.getScript();
```

Ta-da!   CC @vivek  and @michaelneale 

I'm not marking it ready to review *yet* because need to look at it with fresh eyes tomorrow and see if there's a smarter way to implement this.

**TODO:**

- [ ] Rejigger this as [described by Jesse](https://issues.jenkins-ci.org/browse/JENKINS-31582) using a direct serialization of the Step in the Actions for the FlowNode (initial version, avoids some compatibility issues there).   This way we can get at far more info from the execution. 

Catch: this serializes far more data.  It might reduce performance if done without a rewrite of flow node storage.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.1</version>
+            <version>2.6</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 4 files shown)


i386:
Would be nice if steps could expose this easily through their implementations.

One request was for shell, batch and scm steps but undoubtably there will be more steps

michaelneale:
I like it

vivek:
@svanoort This is great but I agree with @i386, in blueocean we will always have to check for such actions vs this is exposed from underlying API.

i386:
Here's a good example of where SCM step descriptions would help the user

nov-23-2016 14-14-25

svanoort:
Closing this because we're going to do this in a better way that is portable across steps, by serializing the step instance itself :)

PR #23, opened 2016-12-29
Robustness improvements

jglick:
- [X] [JENKINS-40613](https://issues.jenkins-ci.org/browse/JENKINS-40613): make sure `FilePath.isDirectory` calls are subject to timeout
- [X] [JENKINS-37730](https://issues.jenkins-ci.org/browse/JENKINS-37730): include more diagnostics in `getStatus`
- [X] [JENKINS-38769](https://issues.jenkins-ci.org/browse/JENKINS-38769): make sure `stop` does something
- [X] [JENKINS-37486](https://issues.jenkins-ci.org/browse/JENKINS-37486): reported `NullPointerException` perhaps due to Guice problems

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -204,14 +204,24 @@ public abstract class DurableTaskStep extends AbstractStepImpl {
             return l.getLogger();
         }
 
-        @Override public void stop(Throwable cause) throws Exception {
+        @Override public void stop(final Throwable cause) throws Exception {
             FilePath workspace = getWorkspace();
             if (workspace != null) {
                 logger().println("Sending interrupt signal to process");
                 LOGGER.log(Level.FINE, "stopping process", cause);
                 controller.stop(workspace, launcher);
+                Timer.get().schedule(new Runnable() {
+                    @Override public void run() {
+                        if (recurrencePeriod > 0) {
+                            recurrencePeriod = 0;
+                            logger().println("After 10s process did not stop");
+                            getContext().onFailure(cause);
+                        }
+                    }
+                }, 10, TimeUnit.SECONDS);
             } else {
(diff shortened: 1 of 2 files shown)


recampbell:
Thanks Jesse. What use-cases can be covered by tests here?

stephenc:
Would be good if you had a test that verified upgrading a system while a durable task is running does not lead to a zombie durable task ;-)

:bee:

jglick:
> a test that verified upgrading a system while a durable task is running does not lead to a zombie durable task

Well, it depends on the nature of the upgrade. The most dangerous changes would be to the serialized form of a `StepExecution` (transitively, so including nested objects such as a `Controller` in this case). This patch should only be touching `transient` fields, though.

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -120,42 +125,50 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
...
+
     }
 
     /**
      * Represents one task that is believed to still be running.
      */
-    @Restricted(NoExternalUse.class)
     @SuppressFBWarnings(value="SE_TRANSIENT_FIELD_NOT_RESTORED", justification="recurrencePeriod is set in onResume, not deserialization")
-    public static final class Execution extends AbstractStepExecutionImpl implements Runnable {
+    static final class Execution extends StepExecution implements Runnable {
```
jglick: Unfortunately, this does seem to cause a `InvalidClassException`:

    Class does not extend stream superclass

Looking to see if there is some way around that which does not require going back to `AbstractStepExecutionImpl`.

jglick: Filed as [JENKINS-40909](https://issues.jenkins-ci.org/browse/JENKINS-40909).

PR #26, opened 2017-01-12
[JENKINS-41010] Expose StepContext from PlaceholderTask

vivek:
[JENKINS-41010](https://issues.jenkins-ci.org/browse/JENKINS-41010)

Exposes underlying StepContext from PlaceholderTask.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -41,6 +41,7 @@ import java.util.logging.Level;
 import static java.util.logging.Level.*;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


vivek:
@jglick that works too. PTAL, thanks.

vivek:
Forgot this:) @reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -41,6 +41,7 @@
 import static java.util.logging.Level.*;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
```
jglick: unused import

jglick:
Occurs to me that the better API would have been an `Action` subtype representing a `FlowNode` which could be attached to a `Queue.Item`. That would avoid the need to depend on implementation classes in the step; you could just go through `workflow-api`.

vivek:
@jglick yeah thought would be cleaner. Ping me if you plan to implement it that way in later version and I can change my code accordingly. Thanks.

PR #27, opened 2017-01-26
[JENKINS-41339] durable-task 1.13 broke an EnvironmentVariablesNodeProperty use case

jglick:
[JENKINS-41339](https://issues.jenkins-ci.org/browse/JENKINS-41339)

Buggy `sh` behavior corrected in `durable-task` 1.13 exposed a bug in the `node` step.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep.java
@@ -24,9 +24,13 @@
 
 package org.jenkinsci.plugins.workflow.steps.durable_task;
 
+import hudson.EnvVars;
 import hudson.Extension;
+import hudson.model.TaskListener;
 import org.jenkinsci.plugins.durabletask.BourneShellScript;
 import org.jenkinsci.plugins.durabletask.DurableTask;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep.java
```diff
@@ -50,6 +54,14 @@ public String getScript() {
         return new BourneShellScript(script);
     }
 
+    @Override public StepExecution start(StepContext context) throws Exception {
+        String path = context.get(EnvVars.class).get("PATH");
+        if (path != null && path.contains("$PATH")) {
+            context.get(TaskListener.class).getLogger().println("Warning: JENKINS-41339 probably bogus PATH=" + path + "; perhaps you meant to use ‘PATH+EXTRA=/something/bin’?");
```
jglick: Not necessary for `EnvironmentVariablesNodeProperty`, which is now fixed, but perhaps for people using `withEnv` incorrectly etc.

stephenc:
:bee:

rsandell:
:bee:

Ericson2314:
Just wanted to say this is a very good best-of-both-worlds fix. Thanks!

PR #29, opened 2017-01-30
[JENKINS-41446] fix $WORKSPACE in ws step

blatinville:
[JENKINS-41446](https://issues.jenkins-ci.org/browse/JENKINS-41446)

I took the EnvStep as example to use the EnvironmentExpander.
As we have only one variable to set, I made ExpanderImpl a bit simpler.

Created unitary tests, including test of NODE_NAME since my first attempt resulted in the variable set to null in the ws step.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
@@ -2,6 +2,10 @@ package org.jenkinsci.plugins.workflow.support.steps;
 
 import com.google.inject.Inject;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.model.Computer;
 import hudson.model.Job;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -60,12 +65,27 @@ public boolean start() throws Exception {
         flowNode.addAction(new WorkspaceActionImpl(workspace, flowNode));
         listener.getLogger().println("Running in " + workspace);
         body = getContext().newBodyInvoker()
-                .withContext(workspace)
+                .withContexts(
+                    EnvironmentExpander.merge(getContext().get(EnvironmentExpander.class),
+                        new ExpanderImpl("WORKSPACE", workspace.getRemote())),
```
jglick: Could even simplify further to include only the path.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -77,9 +77,9 @@ public boolean start() throws Exception {
     private static final class ExpanderImpl extends EnvironmentExpander {
         private static final long serialVersionUID = 1;
         private final Map<String,String> override;
-        private ExpanderImpl(String var, String value) {
+        private ExpanderImpl(String path) {
             this.override = new HashMap<>();
-            this.override.put(var, value);
+            this.override.put("WORKSPACE", path);
```
jglick: What I meant was that you do not need to store a whole `Map` in the serial form—just one `String`.

Not a big deal, but cannot easily be changed once introduced.

blatinville: Sorry, I missed the override method. Just fixed.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -60,12 +65,26 @@ public boolean start() throws Exception {
...
+                .withContexts(
+                    EnvironmentExpander.merge(getContext().get(EnvironmentExpander.class),
+                        new ExpanderImpl(workspace.getRemote())),
+                    workspace)
                 .withCallback(new Callback(lease))
                 .start();
         return false;
     }
 
+    private static final class ExpanderImpl extends EnvironmentExpander {
```
jglick: Probably I need to introduce a convenience implementation of `EnvironmentExpander` that just expands a predefined map of values (not computed on demand), since a few plugins now do this.

PR #30, opened 2017-02-05
[JENKINS-37719] Better recovery from hangs

jglick:
https://github.com/jenkinsci/workflow-cps-plugin/pull/102 ensures an eventual timeout of a build facing a `dockerd` outage ([JENKINS-37719](https://issues.jenkins-ci.org/browse/JENKINS-37719)), even without patching `docker-workflow` itself. This patch also improves upon behavior introduced in #23.

Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/29.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>2.13-20170205.153141-1</version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/29 -->
+            <version>2.13</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -114,7 +114,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>2.13-20170205.153141-1</version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/29 -->
+            <version>2.13</version>
             <classifier>tests</classifier>
             <scope>test</scope>
         </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.11</version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-support</artifactId>
-            <version>2.12</version>
+            <version>2.13-20170205.153141-1</version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/29 -->
```
oleg-nenashev: Still needs to be fixed before the merge

jglick:
@reviewbybees done

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -332,12 +321,10 @@ private void check() {
                             recurrencePeriod = 0;
                             controller.cleanup(workspace);
                         }
-                        return null;
-                    }
-                }, 10, TimeUnit.SECONDS);
             } catch (Exception x) {
                 LOGGER.log(Level.FINE, "could not check " + workspace, x);
                 ws = null;
+                logger().println("Cannot contact " + node + ": " + x); // TODO as above
```
boul3ez: Hi,

We have an issue in our Jenkins: in a pipeline job, a part of a job started on a docker template will never finish if the the machine running the container goes offline. The build remains stuck on this log:
Cannot contact swarm-41aec2277655: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException

My question is : why there is no exception raised here to stop the build and allow to implement a retry strategy?

Any help would be very appreciated.

Thanks.

jglick: Because the node may simply be offline temporarily. Related to [JENKINS-36013](https://issues.jenkins-ci.org/browse/JENKINS-36013).

boul3ez: How to distinguish both cases?
What do you think about adding a configurable timeout to help detect that node is definitely offline (like done at connection I think).

for information, I created the next issue in jenkins bug tracker: https://issues.jenkins-ci.org/browse/JENKINS-43607

PR #32, opened 2017-02-25
Added a Jenkinsfile

jglick:
- [X] add something so that organization folder scanning knows about it
- [x] make sure tests pass on a new baseline
- [x] make sure tests pass on Windows

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin(jenkinsVersions: [null, '2.32.2'])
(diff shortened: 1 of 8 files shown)


jglick:
> create a technical debt issue for the tests without Batch script equivalent

I do not generally bother to use JIRA for non-user-facing technical debt issues like this; existence of `TODO` comments is sufficient reminder to revisit whenever touching related code, or doing bulk cleanups of the plugin, etc.

PR #38, opened 2017-05-09
Exposing NODE_HOME variable in pipeline

arthurvanduynhoven:
Creating/Exposing a variable `NODE_HOME` during use in a pipeline.
This variable is similar to `JENKINS_HOME` but for Jenkins' Agent Nodes.

**Reason/Need for this:**
I have a Jenkins environment that builds same type of projects (Maven, NPM, Composer, etc) using Docker for the build tools. These build tools have their own caching mechanism for their dependencies.
Since using Docker, these containers are short lived and do not persist the cache dependencies. Resulting in the dependencies re-downloading for every build. This is slow and takes up a lot of network bandwidth.

A way to persist the cache dependencies is to mount a shared volume (docker volume mount from host to container) for the dependency cache. To be able to do this we need a writable directory (as the Jenkins user) outside of the workspace (as it may be wiped before each build) so it can be reused during builds.

**Temp/Manual Workaround:**
Login to each Jenkins server (master and agents) and manually create these cache directories in the Jenkins home directory.
Make sure all Jenkins agents have the same user/home directory.

**Workaround Issues:**
Depending on your organization you may not always have authority to create the same setups for Jenkins/home directory (ie Windows vs Unix or system admins different setup procedures).
Extra steps to provision your Jenkins Agents.

**Other Possible Benefits:**
Organizations/users may have Jenkins specific scripts relative to the Jenkins Users home for execution. This provides a better way to get the path instead of doing `../../script.sh` from the workspace.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -539,7 +539,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         }
                         env.put("EXECUTOR_NUMBER", String.valueOf(exec.getNumber()));
                         env.put("NODE_LABELS", Util.join(node.getAssignedLabels(), " "));
-                        env.put("NODE_HOME", node.getRootPath().getRemote());
+                        env.put("NODE_HOME", node.getRootPath() != null ? node.getRootPath().getRemote() : "");
 
                         synchronized (runningTasks) {
                             runningTasks.put(cookie, new RunningTask());
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -539,6 +539,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         }
                         env.put("EXECUTOR_NUMBER", String.valueOf(exec.getNumber()));
                         env.put("NODE_LABELS", Util.join(node.getAssignedLabels(), " "));
+                        env.put("NODE_HOME", node.getRootPath().getRemote());
 
                         synchronized (runningTasks) {
                             runningTasks.put(cookie, new RunningTask());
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -74,6 +74,30 @@
         r.assertLogContains("My name on a slave is node-test using labels fast node-test unix", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
+    /**
+     * Verifies if NODE_HOME environment variable is available on a slave node and on master.
```
oleg-nenashev: s/"slave node"/agent

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -74,6 +74,30 @@
...
+     *
+     * @throws Exception
```
oleg-nenashev: If you declare it, please document to avoid the Javadoc warning

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -539,6 +539,7 @@ private static void finish(@CheckForNull final String cookie) {
                         }
                         env.put("EXECUTOR_NUMBER", String.valueOf(exec.getNumber()));
                         env.put("NODE_LABELS", Util.join(node.getAssignedLabels(), " "));
+                        env.put("NODE_HOME", node.getRootPath().getRemote());
```
oleg-nenashev: FindBugs will fail on it, `getRootPath()` is CheckForNull

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
arthurvanduynhoven: done.

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
arthurvanduynhoven: done.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
arthurvanduynhoven: Done - added additional null check.

jglick:
> we need a writable directory (as the Jenkins user) outside of the workspace (as it may be wiped before each build) so it can be reused during builds

```groovy
def tmpdir = pwd tmp:true
```

PR #39, opened 2017-05-10
Fix JENKINS-28182

mmitche:
The durable task plugin should use a different environment variable to track child processes.  Originally it overrode the JENKINS_SERVER_COOKIE variable, to ensure that the process killing done at the end of a node block would not kill tasks running in other node blocks on the same machine.  However, the killer would attempt to kill processes based on the original value of JENKINS_SERVER_COOKIE.  This means that no processes started in a node block would be killed.  With this fix, JENKINS_NODE_COOKIE is now used.  Added test to ensure that killing functionality works properly.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -212,7 +212,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         @Nullable Launcher launcher;
     }
 
-    private static final String COOKIE_VAR = "JENKINS_SERVER_COOKIE";
+    private static final String COOKIE_VAR = "JENKINS_NODE_COOKIE";
 
     @ExportedBean
     public static final class PlaceholderTask implements ContinuedTask, Serializable, AccessControlled {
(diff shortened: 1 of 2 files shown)


mmitche:
Fixes https://issues.jenkins-ci.org/browse/JENKINS-28182

mmitche:
@oleg-nenashev Are you sure?  In the current implementation, nothing is killed anyway (because cookie var's value never matches any process's environment).

mmitche:
@jglick Is there any thought about implementing a pipeline step that calls the process cleanup explicitly?

The scenario that I'm seeing occasionally is:
* We want to clean the workspace after using a node {}
* A process inside the workspace does not exit
* WsCleanup called at the end will fail, since the process killer has not run.

Ideally, we would allow for workspace cleanup post-release of node, or allow a specific build step which calls this process killer.

jzylkin:
Question: If I have two node blocks executing on the same node machine, and both of them spawn processes, how do I ensure that the process killer does not kill both processes when only one of the two node blocks is exited?

mmitche:
This is what this fixes.  When the node enters, it gets a cookie specific to that node block, not the overall connection.

jzylkin:
So, in a Jenkins pipeline, if I create two node{} blocks, and use an sh step within each one, they will spawn and destroy processes independently?  Or do I have to invoke this specific plugin (durable processes) specifically somehow (In which case I don't know how)?

mmitche:
This plugin defines the node block itself.  And yes, it will destroy the processes independently.   If you don't see the behavior, let us know.

jzylkin:
Thanks very much for your response. I do see now that the COOKIE values are unique for each node, regardless of what machine they are on.  

I am trying to track down an issue where my bourne shell processes mysteriously exit at random times occasionally, with exit code "2" (whatever that means) and no other errors in the log. Wish me luck.

PR #42, opened 2017-06-19
[JENKINS-44636] Add ExecutorTaskInfoAction for node status info

abayer:
[JENKINS-44636](https://issues.jenkins-ci.org/browse/JENKINS-44636)

Initial work on adding a new action, currently named `ExecutorTaskInfoAction`, to go on the initial `StepStartNode`s for `ExecutorStep`s. It'll record why the `PlaceholderTask` is still in queue (updating at the same time that we currently log the cause of blockage to the build log), the name of the agent it's running on once it's allocated, and when the `PlaceholderTask` left the queue, either to start executing or due to cancellation. It's also exposing `isRunning()`, `isQueued()`, and `isCanceled()` booleans.

This is just the first part of JENKINS-44646 - presumably Blue Ocean will need to start looking for `ExecutorTaskInfoAction`, via Bismuth, I guess?

upstream of https://github.com/jenkinsci/workflow-api-plugin/pull/42

cc @reviewbybees @i386


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorTaskInfoAction.java
```diff
@@ -0,0 +1,88 @@
...
+
+    void setAgent(@Nonnull FilePath workspace) {
+        // Because we're not blocked any more at this point!
+        this.whyBlocked = null;
+        this.whenStartedOrCanceled = System.currentTimeMillis();
+        this.agent = FilePathUtils.getNodeName(workspace);
+    }
+
+    @CheckForNull
+    public String getAgent() {
```
jglick: This is already handled by `WorkspaceAction`.

abayer: Yeah, was debating whether to duplicate it here or not.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -85,10 +86,25 @@
...
-        if (Queue.getInstance().schedule2(task, 0).getCreateItem() == null) {
+        Queue.Item taskInQueue = Queue.getInstance().schedule2(task, 0).getCreateItem();
+        if (taskInQueue == null) {
             // There can be no duplicates. But could be refused if a QueueDecisionHandler rejects it for some odd reason.
             throw new IllegalStateException("failed to schedule task");
         }
+        String initialWhyBlocked = taskInQueue.getWhy();
+        if (initialWhyBlocked == null) {
+            // Reuse the general
+            initialWhyBlocked = hudson.model.Messages.Queue_Unknown();
```
jglick: Huh? If it is not blocked, why are we claiming it is?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -85,10 +86,25 @@
...
+        }
+
+        ExecutorTaskInfoAction taskAction = flowNode.getAction(ExecutorTaskInfoAction.class);
+        if (taskAction == null) {
+            // Populate the initial version of the ExecutorTaskInfoAction.
+            flowNode.addAction(new ExecutorTaskInfoAction(initialWhyBlocked, flowNode));
+        } else {
```
jglick: How could it exist already if we are in `start`?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -197,7 +227,22 @@ public void stop(Throwable cause) {
         @Override public void onLeft(Queue.LeftItem li) {
             if (li.isCancelled()) {
                 if (li.task instanceof PlaceholderTask) {
-                    (((PlaceholderTask) li.task).context).onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));
+                    PlaceholderTask p = (PlaceholderTask)li.task;
+                    p.context.onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));
+                    try {
+                        FlowNode n = p.getNode();
```
jglick: careful, this could block

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -197,7 +227,22 @@ public void stop(Throwable cause) {
...
+                        if (n != null) {
+                            ExecutorTaskInfoAction t = n.getAction(ExecutorTaskInfoAction.class);
+                            if (t == null) {
+                                t = new ExecutorTaskInfoAction(n);
```
jglick: But not attaching it?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Because it is in the queue, however briefly - this'll get updated away as soon as it launches.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: `onResume()` potentially calls `start()` - I decided to play it safe for that scenario.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Goddangit, I forgot about that. Lemme think about that.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Whoopsie.

abayer:
I have no idea why compilation is failing. Sigh.

i386:
> This is just the first part of JENKINS-44646 - presumably Blue Ocean will need to start looking for ExecutorTaskInfoAction, via Bismuth, I guess?

@abayer I assume so - via some thin wrapper for steps and their statuses?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -91,10 +92,23 @@
     @Override
     public boolean start() throws Exception {
         final PlaceholderTask task = new PlaceholderTask(getContext(), step.getLabel(), run);
-        if (Queue.getInstance().schedule2(task, 0).getCreateItem() == null) {
+        Queue.Item taskInQueue = Queue.getInstance().schedule2(task, 0).getCreateItem();
+        if (taskInQueue == null) {
             // There can be no duplicates. But could be refused if a QueueDecisionHandler rejects it for some odd reason.
             throw new IllegalStateException("failed to schedule task");
```
svanoort: I'm more than a bit twitchy about the throwing of IllegalStateException -- won't that trigger a failure of the flow?  Maybe better to just `log` a complaint?

This sort of assumption has bitten me before.

abayer: This isn't a change of behavior, though, just a change of syntax - I'm just keeping the `taskInQueue` around for later use, otherwise it's the same as it was before in re: the `IllegalStateException` - if `...getCreateItem()` is `null`, it'll throw.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -244,10 +244,16 @@ public void stop(Throwable cause) {
          */
         private String cookie;
 
+        /**
+         * Unique cookie set when the task is queued. Used for status lookups.
+         */
+        private final String queueCookie;
```
svanoort: I think this can be null if deserialized or loading an old PlaceHolderTask -- we should be using a synchronized getter that creates the queueCookie if null.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -637,7 +647,9 @@ private static void finish(@CheckForNull final String cookie) {
                         // Cf. AbstractBuild.getEnvironment:
                         env.put("WORKSPACE", workspace.getRemote());
                         FlowNode flowNode = context.get(FlowNode.class);
-                        flowNode.addAction(new WorkspaceActionImpl(workspace, flowNode));
+                        if (flowNode != null) {
```
jglick: Huh? This should never be null. This can be reverted.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -240,10 +244,16 @@ public void stop(Throwable cause) {
          */
         private String cookie;
 
+        /**
+         * Unique cookie set when the task is queued. Used for status lookups.
```
jglick: No, do not do this. Use `Queue.Item.id` (bearing serial compatibility in mind).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -95,6 +96,8 @@ public boolean start() throws Exception {
             // There can be no duplicates. But could be refused if a QueueDecisionHandler rejects it for some odd reason.
             throw new IllegalStateException("failed to schedule task");
         }
+        flowNode.addAction(new TaskInfoActionImpl(task.queueCookie));
```
jglick: Use the return value of `schedule2` here.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -203,7 +206,8 @@ public void stop(Throwable cause) {
         @Override public void onLeft(Queue.LeftItem li) {
             if (li.isCancelled()) {
                 if (li.task instanceof PlaceholderTask) {
-                    (((PlaceholderTask) li.task).context).onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));
+                    PlaceholderTask p = (PlaceholderTask)li.task;
+                    p.context.onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));
```
jglick: Gratuitous diff hunk?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -726,6 +738,29 @@ public String getUrl() {
         private static final long serialVersionUID = 1098885580375315588L; // as of 2.12
     }
 
+    private static final class TaskInfoActionImpl extends TaskInfoAction {
+        /**
+         * Used to identify the task in the queue, so that its status can be identified.
+         */
+        private String taskQueueCookie;
```
jglick: `long id`

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -726,6 +738,29 @@ public String getUrl() {
...
+
+        TaskInfoActionImpl(String taskQueueCookie) {
+            this.taskQueueCookie = taskQueueCookie;
+        }
+
+        @Override
+        @CheckForNull
+        protected Queue.Item itemInQueue() {
+            for (Queue.Item item : Queue.getInstance().getItems()) {
```
jglick: `Queue.getItem(long)`

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Looks to be a lurking remnant of an earlier approach. Cleaning up.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Ah-ha! Thanks!

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: In practice, it's never null, but since `context.get` isn't `Nonnull`, my IDE is convinced it *could* be null, so I prefer to shut up the IDE and play it safe. =)

jglick: It is `@Nullable`, as it should be. What is wrong with your IDE?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -203,8 +207,7 @@ public void stop(Throwable cause) {
         @Override public void onLeft(Queue.LeftItem li) {
             if (li.isCancelled()) {
                 if (li.task instanceof PlaceholderTask) {
-                    (((PlaceholderTask) li.task).context).onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));
-                }
+                    (((PlaceholderTask) li.task).context).onFailure(new AbortException(Messages.ExecutorStepExecution_queue_task_cancelled()));                }
```
jglick: Revert.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -240,10 +243,16 @@ public void stop(Throwable cause) {
          */
         private String cookie;
 
+        /**
+         * Unique cookie set when the task is queued. Used for status lookups.
+         */
+        private final String queueCookie;
```
jglick: No longer used; delete.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
abayer: Whoops!

PR #44, opened 2017-07-27
Amending JENKINS-34542 fix for JENKINS-45553

jglick:
Amending #3 for [JENKINS-45553](https://issues.jenkins-ci.org/browse/JENKINS-45553). Otherwise we tend to suck up all available `Timer` threads, causing various problems such as `copyLogs` starvation (pending JENKINS-38381).

@reviewbybees


jglick:
@reviewbybees done

PR #47, opened 2017-08-15
Fix [JENKINS-36013] by adding a timeout if waiting to resume on unknown node

svanoort:
See [JENKINS-36013](https://issues.jenkins-ci.org/browse/JENKINS-36013). 

For this I've not limited this to EphemeralNodes because (for example) Docker slaves may not implement EphemeralNode but may exhibit the same "once it's gone it ain't coming back" behavior.  

However I did add a boolean to track if the slave is ephemeral so that goes into the logs.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -126,7 +126,7 @@ public class ExecutorPickle extends Pickle {
                                 Queue.getInstance().cancel(item);
                                 if (isEphemeral()) {
                                     throw new AbortException(MessageFormat.format("Killed {0} because EphemeralNode {1} is never going to reappear, by definition!",
-                                            new Object[]{item, TIMEOUT_WAITING_FOR_NODE_MILLIS, task.getAssignedLabel().toString()}));
+                                            new Object[]{item, task.getAssignedLabel().toString()}));
                                 } else {
                                     throw new AbortException(MessageFormat.format("Killed {0} after waiting for {1} ms because we assume unknown Node {1} is never going to appear!",
                                             new Object[]{item, TIMEOUT_WAITING_FOR_NODE_MILLIS, placeholder.getAssignedLabel().toString()}));
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -28,6 +28,7 @@ import com.google.common.util.concurrent.ListenableFuture;
 import hudson.AbortException;
 import hudson.Extension;
 import hudson.model.Executor;
+import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.OneOffExecutor;
 import hudson.model.Queue;
@@ -37,9 +38,12 @@ import hudson.model.queue.SubTask;
 import java.util.concurrent.Future;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.pickles.Pickle;
 import org.jenkinsci.plugins.workflow.steps.durable_task.Messages;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution;
 
 /**
  * Persists an {@link Executor} as the {@link hudson.model.Queue.Task} it was running.
@@ -75,7 +79,7 @@ public class ExecutorPickle extends Pickle {
             @Override
             protected Executor tryResolve() throws Exception {
                 Queue.Item item;
-                if (itemID == 0) {
+                if (itemID == 0) { // Not scheduled yet
                     item = Queue.getInstance().schedule2(task, 0).getItem();
                     if (item == null) {
                         // TODO should also report when !ScheduleResult.created, since that is arguably an error
@@ -93,8 +97,20 @@ public class ExecutorPickle extends Pickle {
                 Future<Queue.Executable> future = item.getFuture().getStartCondition();
 
                 if (!future.isDone()) {
-                    // TODO JENKINS-26130 we might be able to detect that the item is blocked on an agent which has been deleted (not just offline), and abort ourselves
-                    LOGGER.log(Level.FINER, "{0} not yet started", item);
+                    Queue.Task task = item.task;
+                    Label nodeNameLabel = task.getAssignedLabel();  // Once started, this is set to the Computer name
+                    // See javadocs for ExecutorStepExecution.PlaceholderTask.label
+                    // TODO see if we need to do something with the ExecutorStepExecution.PlaceholderTask.runningTasks
+                    if (nodeNameLabel != null && task.getLastBuiltOn() == null) {  // Can't find node it was built on!
+                        if (task instanceof ExecutorStepExecution.PlaceholderTask) {
+                            Queue.getInstance().cancel(item);
+                            LOGGER.log(Level.INFO, "{0} tried to resume on unknown Node {1}, assuming that the Node isn't coming back and aborting the build ",
+                                    new Object[]{item, nodeNameLabel.toString()});
+                        }
+                    } else {
+                        // TODO JENKINS-26130 we might be able to detect that the item is blocked on an agent which has been deleted (not just offline), and abort ourselves
+                        LOGGER.log(Level.FINER, "{0} not yet started", item);
+                    }
                     return null;
                 }
 
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -35,10 +35,14 @@ import hudson.model.Queue;
 import hudson.model.TaskListener;
 import hudson.model.queue.CauseOfBlockage;
 import hudson.model.queue.SubTask;
+
+import java.text.MessageFormat;
 import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import hudson.slaves.EphemeralNode;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
 import org.jenkinsci.plugins.workflow.pickles.Pickle;
@@ -57,6 +61,10 @@ public class ExecutorPickle extends Pickle {
 
     private final Queue.Task task;
 
+    private final boolean isEphemeral;
+
+    static long TIMEOUT_WAITING_FOR_EPHMERAL_NODE = TimeUnit.MINUTES.toMillis(5);
+
     private ExecutorPickle(Executor executor) {
         if (executor instanceof OneOffExecutor) {
             throw new IllegalArgumentException("OneOffExecutor not currently supported");
@@ -65,6 +73,7 @@ public class ExecutorPickle extends Pickle {
         if (exec == null) {
             throw new IllegalArgumentException("cannot save an Executor that is not running anything");
         }
+        this.isEphemeral = executor.getOwner().getNode() instanceof EphemeralNode;
         SubTask parent = exec.getParent();
         this.task = parent instanceof Queue.Task ? (Queue.Task) parent : parent.getOwnerTask();
         if (task instanceof Queue.TransientTask) {
@@ -73,9 +82,15 @@ public class ExecutorPickle extends Pickle {
         LOGGER.log(Level.FINE, "saving {0}", task);
     }
 
+    public boolean isEphemeral() {
+        return isEphemeral;
+    }
+
     @Override public ListenableFuture<Executor> rehydrate(final FlowExecutionOwner owner) {
         return new TryRepeatedly<Executor>(1, 0) {
             long itemID;
+            long startTimeMillis = System.currentTimeMillis();
+
             @Override
             protected Executor tryResolve() throws Exception {
                 Queue.Item item;
@@ -98,18 +113,19 @@ public class ExecutorPickle extends Pickle {
 
                 if (!future.isDone()) {
                     Queue.Task task = item.task;
-                    Label nodeNameLabel = task.getAssignedLabel();  // Once started, this is set to the Computer name
+                    Label nodeNameLabel = task.getAssignedLabel();  // Once execution started, this was set to the Computer name
                     // See javadocs for ExecutorStepExecution.PlaceholderTask.label
+                    long waitTime = System.currentTimeMillis()-startTimeMillis;
                     // TODO see if we need to do something with the ExecutorStepExecution.PlaceholderTask.runningTasks
                     if (nodeNameLabel != null && task.getLastBuiltOn() == null) {  // Can't find node it was built on!
-                        if (task instanceof ExecutorStepExecution.PlaceholderTask) {
+                        if (waitTime > TIMEOUT_WAITING_FOR_EPHMERAL_NODE && task instanceof ExecutorStepExecution.PlaceholderTask) {
                             Queue.getInstance().cancel(item);
-                            LOGGER.log(Level.INFO, "{0} tried to resume on unknown Node {1}, assuming that the Node isn't coming back and aborting the build ",
-                                    new Object[]{item, nodeNameLabel.toString()});
+                            throw new AbortException(MessageFormat.format("Killed {0} after waiting for {1} millis because we assume unknown Node {1} (ephemeral={2}) is never going to reappear!",
+                                    new Object[]{item, waitTime, nodeNameLabel.toString(), isEphemeral()}));
                         }
                     } else {
                         // TODO JENKINS-26130 we might be able to detect that the item is blocked on an agent which has been deleted (not just offline), and abort ourselves
-                        LOGGER.log(Level.FINER, "{0} not yet started", item);
+                        LOGGER.log(Level.FINER, "{0} not yet started after waiting for {1} millis", new Object[]{item, waitTime});
                     }
                     return null;
                 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -44,6 +44,7 @@ import java.util.logging.Logger;
 
 import hudson.slaves.EphemeralNode;
 import jenkins.model.Jenkins;
+import jenkins.util.SystemProperties;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -53,6 +61,10 @@
 
     private final Queue.Task task;
 
+    private final boolean isEphemeral;
```
svanoort: Will be curious to see what others think of this (or if they think it's not worth tracking).

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -237,7 +238,7 @@ public DescriptorImpl() {
     @Test public void deadStep() throws Exception {
         logging.record(DurableTaskStep.class, Level.INFO).record(CpsStepContext.class, Level.INFO).capture(100);
         WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
-        p.setDefinition(new CpsFlowDefinition("try {node {isUnix() ? sh('sleep infinity') : bat('ping 127.0.0.1 > nul')}} catch (e) {sleep 1; throw e}", true));
+        p.setDefinition(new CpsFlowDefinition("try {node {isUnix() ? sh('sleep 1000000') : bat('ping 127.0.0.1 > nul')}} catch (e) {sleep 1; throw e}", true));
```
svanoort: Mac doesn't support 'sleep infinity' ;)

svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
jglick: Definitely you need to track this. If the node was ephemeral, you need not bother waiting at all—it cannot come back, since the very record of its existence has been discarded already.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -53,6 +61,10 @@
...
+
+    static long TIMEOUT_WAITING_FOR_EPHMERAL_NODE = TimeUnit.MINUTES.toMillis(5);
```
jglick: Typo. Also probably this should use `SystemProperties` (2.4+, but you ought to be on 2.7.x anyway) to let it be configurable at runtime.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -69,13 +82,19 @@ private ExecutorPickle(Executor executor) {
...
     }
 
+    public boolean isEphemeral() {
+        return isEphemeral;
+    }
+
     @Override public ListenableFuture<Executor> rehydrate(final FlowExecutionOwner owner) {
         return new TryRepeatedly<Executor>(1, 0) {
             long itemID;
+            long startTimeMillis = System.currentTimeMillis();
```
jglick: Use `System.nanoTime` for measuring elapsed time.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -93,8 +112,21 @@ protected Executor tryResolve() throws Exception {
                 Future<Queue.Executable> future = item.getFuture().getStartCondition();
 
                 if (!future.isDone()) {
-                    // TODO JENKINS-26130 we might be able to detect that the item is blocked on an agent which has been deleted (not just offline), and abort ourselves
-                    LOGGER.log(Level.FINER, "{0} not yet started", item);
+                    Queue.Task task = item.task;
+                    Label nodeNameLabel = task.getAssignedLabel();  // Once execution started, this was set to the Computer name
```
jglick: Hmm. That is true of `PlaceholderTask`. But `ExecutorPickle` was originally designed to potentially support other `Executor`s. So it would be better to do the `instanceof` check to begin with, and for any other kind of `Task`, just keep the original behavior.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -93,8 +112,21 @@ protected Executor tryResolve() throws Exception {
...
+                    // See javadocs for ExecutorStepExecution.PlaceholderTask.label
+                    long waitTime = System.currentTimeMillis()-startTimeMillis;
+                    // TODO see if we need to do something with the ExecutorStepExecution.PlaceholderTask.runningTasks
+                    if (nodeNameLabel != null && task.getLastBuiltOn() == null) {  // Can't find node it was built on!
```
jglick: This is a roundabout way of checking things. `getAssignedLabel()` may or may not be null prior to the start of the body: depends on whether a label was passed to the `node` step. And the same is true of `getLastBuiltOn()`: if a self-label was passed as an explicit argument, and the node is still around, it will be non-null even if the task has not yet been scheduled to an executor. This logic may be right but it is hard for me to say for sure—think it out carefully.

BTW if you need to know for sure whether the body has started, you can check `PlaceholderTask.cookie`.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -93,8 +112,21 @@ protected Executor tryResolve() throws Exception {
...
+                        if (waitTime > TIMEOUT_WAITING_FOR_EPHMERAL_NODE && task instanceof ExecutorStepExecution.PlaceholderTask) {
+                            Queue.getInstance().cancel(item);
+                            throw new AbortException(MessageFormat.format("Killed {0} after waiting for {1} millis because we assume unknown Node {1} (ephemeral={2}) is never going to reappear!",
```
jglick: `Util.getTimeSpanString`

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -93,8 +112,21 @@ protected Executor tryResolve() throws Exception {
...
+                                    new Object[]{item, waitTime, nodeNameLabel.toString(), isEphemeral()}));
+                        }
+                    } else {
+                        // TODO JENKINS-26130 we might be able to detect that the item is blocked on an agent which has been deleted (not just offline), and abort ourselves
```
jglick: This comment can be deleted now, as it was referring to the exact condition you purport to fix in this PR!

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
jglick: Ah, GNU extension I suppose.

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -107,4 +124,99 @@
...
+            }
+        });
+
+        r.addStep(new Statement() {
+            // Start up a build and then reboot and take the node offline
+            @Override public void evaluate() throws Throwable {
+                assertNull(r.j.jenkins.getNode("ghostly")); // Make sure test impl is correctly deleted
+                WorkflowRun run = r.j.jenkins.getItemByFullName("p", WorkflowJob.class).getLastBuild();
+                r.j.waitForMessage("Waiting to resume", run);
+                Thread.sleep(1000L);
```
jglick: Why?

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -107,4 +124,99 @@
...
+                Assert.assertEquals("Queue should still have single build Item waiting to resume but didn't", 1, Queue.getInstance().getItems().length);
+
+                try {
+                    Thread.sleep(ExecutorPickle.TIMEOUT_WAITING_FOR_EPHMERAL_NODE + 1000L);
+                    Assert.assertEquals("Should have given up and killed the Task representing the resuming build", 0, Queue.getInstance().getItems().length );
+                    Assert.assertFalse(run.isBuilding());
+                    r.j.assertBuildStatus(Result.FAILURE, run);
+                    Assert.assertEquals(0, r.j.jenkins.getQueue().getItems().length);
+                } catch (InterruptedIOException ioe) {
+                    Assert.fail("Waited for build to detect loss of ephemeral node and it didn't!");
```
jglick: Unnecessary—an uncaught exception in a test will be treated as a failure.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
svanoort: I reworked this a bit and added some comments, though it's still going to be a bit cryptic.

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
michaelneale: the failure is slightly more obvious this way though...

michaelneale:
build failure looks unrelated to changes - terminated prematurely. cosmic rays etc.

svanoort:
That's a legit failure -- but in the testcase itself.  Fixed that,  but hit a bug somewhere in the way agents interact with the RestartableJenkinsRule which I'm slapping down now.

michaelneale:
@jglick look ok to you now?

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
jtnord: This is not correct and likely to not be correct on windows if you actually expect it to do something for some amount of time.

`ping` on windows does not ping indefinatly but sends 4 pings 1 second (or so) apart by default.

```
C:\Users\jnord>ping 127.0.0.1

Pinging 127.0.0.1 with 32 bytes of data:
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128
Reply from 127.0.0.1: bytes=32 time<1ms TTL=128

Ping statistics for 127.0.0.1:
    Packets: Sent = 4, Received = 4, Lost = 0 (0% loss),
Approximate round trip times in milli-seconds:
    Minimum = 0ms, Maximum = 0ms, Average = 0ms

C:\Users\jnord>ping /?

Usage: ping [-t] [-a] [-n count] [-l size] [-f] [-i TTL] [-v TOS]
            [-r count] [-s count] [[-j host-list] | [-k host-list]]
            [-w timeout] [-R] [-S srcaddr] [-c compartment] [-p]
            [-4] [-6] target_name

Options:
    -t             Ping the specified host until stopped.
                   To see statistics and continue - type Control-Break;
                   To stop - type Control-C.
    -a             Resolve addresses to hostnames.
    -n count       Number of echo requests to send.
    -l size        Send buffer size.
    -f             Set Don't Fragment flag in packet (IPv4-only).
    -i TTL         Time To Live.
    -v TOS         Type Of Service (IPv4-only. This setting has been deprecated
                   and has no effect on the type of service field in the IP
                   Header).
    -r count       Record route for count hops (IPv4-only).
    -s count       Timestamp for count hops (IPv4-only).
    -j host-list   Loose source route along host-list (IPv4-only).
    -k host-list   Strict source route along host-list (IPv4-only).
    -w timeout     Timeout in milliseconds to wait for each reply.
    -R             Use routing header to test reverse route also (IPv6-only).
                   Per RFC 5095 the use of this routing header has been
                   deprecated. Some systems may drop echo requests if
                   this header is used.
    -S srcaddr     Source address to use.
    -c compartment Routing compartment identifier.
    -p             Ping a Hyper-V Network Virtualization provider address.
    -4             Force using IPv4.
    -6             Force using IPv6.
```

michaelneale: For some reason I still remember this and am always surprised when a *nix ping doesn't stop after 4

svanoort: @jtnord @michaelneale -- Though neither is actually related to the change I'm making, may as well clean that up too when fixing the other test bug that had been lurking in this code.  Done.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -89,7 +90,7 @@ public boolean isEphemeral() {
     @Override public ListenableFuture<Executor> rehydrate(final FlowExecutionOwner owner) {
         return new TryRepeatedly<Executor>(1, 0) {
             long itemID;
-            long startTimeMillis = System.currentTimeMillis();
+            long endTimeNanos = System.nanoTime() + TIMEOUT_WAITING_FOR_NODE_MILLIS*1000000;
```
jglick:
```java
* 1_000_000
```

might be more legible

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -113,20 +114,27 @@ protected Executor tryResolve() throws Exception {
...
+                        // Non-null cookie means body has begun execution, i.e. we started using a node
+                        // PlaceholderTask#getAssignedLabel is set to the Node name when execution starts
+                        // Thus we're guaranteeing the execution began and the Node is now unknown.
+                        // Theoretically it's safe to simply fail earlier when rehydrating any EphemeralNode... but we're being extra safe.
+                        if (placeholder.getCookie() != null && Jenkins.getActiveInstance().getNode(placeholder.getAssignedLabel().getName()) == null ) {
+                            if (isEphemeral() || System.nanoTime() > endTimeNanos) {
+                                Queue.getInstance().cancel(item);
+                                if (isEphemeral()) {
+                                    throw new AbortException(MessageFormat.format("Killed {0} because EphemeralNode {1} is never going to reappear, by definition!",
+                                            new Object[]{item, task.getAssignedLabel().toString()}));
```
jglick: BTW `Queue.Item.toString()` is not going to be particularly user-friendly; will look something like

    hudson.model.Queue$WaitingItem:ExecutorStepExecution.PlaceholderTask{runId=folder/jobname#123,label=,context=CpsStepContext[23:node]:Owner[folder/jobname/123:folder/jobname #123],cookie=dead-beef-1234-5678}

Whew!

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -113,20 +114,27 @@ protected Executor tryResolve() throws Exception {
...
+                                } else {
+                                    throw new AbortException(MessageFormat.format("Killed {0} after waiting for {1} ms because we assume unknown Node {1} is never going to appear!",
+                                            new Object[]{item, TIMEOUT_WAITING_FOR_NODE_MILLIS, placeholder.getAssignedLabel().toString()}));
```
jglick: `Util.getTimeSpanString`

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -238,7 +238,7 @@ public DescriptorImpl() {
     @Test public void deadStep() throws Exception {
         logging.record(DurableTaskStep.class, Level.INFO).record(CpsStepContext.class, Level.INFO).capture(100);
         WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "p");
-        p.setDefinition(new CpsFlowDefinition("try {node {isUnix() ? sh('sleep 1000000') : bat('ping 127.0.0.1 > nul')}} catch (e) {sleep 1; throw e}", true));
+        p.setDefinition(new CpsFlowDefinition("try {node {isUnix() ? sh('sleep 1000000') : bat('ping -t 127.0.0.1 > nul')}} catch (e) {sleep 1; throw e}", true));
```
jglick: Unrelated flake I guess?

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -67,8 +74,30 @@
 
     @ClassRule public static BuildWatcher buildWatcher = new BuildWatcher();
     @Rule public RestartableJenkinsRule r = new RestartableJenkinsRule();
+    @Rule public TemporaryFolder tmp = new TemporaryFolder();
     //@Rule public LoggerRule logging = new LoggerRule().record(Queue.class, Level.FINE);
 
+    private static Process jnlpProc;
+    private void startJnlpProc() throws Exception {
```
jglick: Ideally would be shared from, IIRC, `ShellStepTest`.

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -67,8 +74,30 @@
...
+        killJnlpProc();
+        ProcessBuilder pb = new ProcessBuilder(JavaEnvUtils.getJreExecutable("java"), "-jar", Which.jarFile(Launcher.class).getAbsolutePath(), "-jnlpUrl", r.j.getURL() + "computer/ghostly/slave-agent.jnlp");
+        try {
+            ProcessBuilder.class.getMethod("inheritIO").invoke(pb);
+        } catch (NoSuchMethodException x) {
+            // prior to Java 7
```
jglick: Can be simplified, we are on `java.level` 7 now, right?

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -172,7 +201,7 @@ public Node asNode() {
 
     /**
      * Test that {@link ExecutorPickle} won't spin forever trying to rehydrate if it was using an
-     *  EphemeralNode that disappeared and will never reappear... but still waits a little bit to find out.
+     *  EphemeralNode that disappeared and will never reappear... for EphemeralNodes it should just die.
```
jglick: BTW `@link`

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
svanoort: @jglick It appears this behavior is very much unsafe due to the use of the Swarm Plugin -- please see https://issues.jenkins-ci.org/browse/JENKINS-36013?focusedCommentId=311834&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-311834  -- we might argue that perhaps this is not a correct use of EphemeralNode but in any case the fix in DurableTask is trivial.

PR #48, opened 2017-08-26
Make ExecutorPickle wait on EphemeralNodes when trying to resume, restoring original behavior of fix to JENKINS-36013

svanoort:
Per comment here, immediately failing builds with an ExecutorPickle that depends on an EphemeralNode is not safe to do, due to the Swarm Plugin.   Documented [in the JIRA comments](https://issues.jenkins-ci.org/browse/JENKINS-36013?focusedCommentId=311834&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-311834).

So, I am restoring my original behavior (apply a timeout in all cases).  Can't remove the new field, though.

@reviewbybees CC @jglick

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -122,15 +122,10 @@ public class ExecutorPickle extends Pickle {
                         // Thus we're guaranteeing the execution began and the Node is now unknown.
                         // Theoretically it's safe to simply fail earlier when rehydrating any EphemeralNode... but we're being extra safe.
                         if (placeholder.getCookie() != null && Jenkins.getActiveInstance().getNode(placeholder.getAssignedLabel().getName()) == null ) {
-                            if (isEphemeral() || System.nanoTime() > endTimeNanos) {
+                            if (System.nanoTime() > endTimeNanos) {
                                 Queue.getInstance().cancel(item);
-                                if (isEphemeral()) {
-                                    throw new AbortException(MessageFormat.format("Killed {0} because EphemeralNode {1} is never going to reappear, by definition!",
-                                            new Object[]{item, task.getAssignedLabel().toString()}));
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -122,15 +122,10 @@ protected Executor tryResolve() throws Exception {
                         // Thus we're guaranteeing the execution began and the Node is now unknown.
                         // Theoretically it's safe to simply fail earlier when rehydrating any EphemeralNode... but we're being extra safe.
                         if (placeholder.getCookie() != null && Jenkins.getActiveInstance().getNode(placeholder.getAssignedLabel().getName()) == null ) {
-                            if (isEphemeral() || System.nanoTime() > endTimeNanos) {
+                            if (System.nanoTime() > endTimeNanos) {
                                 Queue.getInstance().cancel(item);
-                                if (isEphemeral()) {
```
jglick: Then you might as well delete the `ephemeral` flag.

svanoort:
@jglick The exact required behavior of EphemeralNodes is pretty poorly defined - the node itself may not be persistent with regard to the master but there's nothing to prevent it from being recreated once removed (as Swarm plugin does when nodes reconnect).   

The safest thing to do is to ensure we're not prematurely terminating builds and just use the original grace period -- there's no guarantee there are no other oddball implementations.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
svanoort: I'd like to keep that info handy for future use, plus it avoids grumping from XStream because the field no longer exists.

jglick:
Suggest developing automated tests in `swarm-plugin` capable of reproducing the original behavior, the regression, and the fix.

jglick:
Amends #47 FTR.

PR #51, opened 2017-09-21
[JENKINS-46738] Deadlock in PlaceholderExecutable

jglick:
[JENKINS-46738](https://issues.jenkins-ci.org/browse/JENKINS-46738)

@reviewbybees


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -688,16 +688,20 @@ public void run() {
                             }
                             LOGGER.log(FINE, "interrupted {0}", cookie);
                             // TODO save the BodyExecution somehow and call .cancel() here; currently we just interrupt the build as a whole:
-                            Executor masterExecutor = r.getExecutor();
```
jglick: [diff for reviewers](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/51/files?w=1)

PR #53, opened 2017-09-27
Use a dedicated thread pool rather than Timer.get for DurableTaskStep.Execution.check

jglick:
Analogous to #44. Downstream of https://github.com/jenkinsci/workflow-support-plugin/pull/46.

@reviewbybees


jglick:
Very flaky tests. Looking into it.

jglick:
@reviewbybees done

PR #56, opened 2017-11-15
Make sure we are testing against JENKINS-47791

jglick:
Pick up https://github.com/jenkinsci/durable-task-plugin/pull/49 in our tests here.

@reviewbybees


jglick:
@reviewbybees done

PR #58, opened 2017-11-26
PlaceHolderTask should be submitted with a reference to the upstream Run.

oleg-nenashev:
I have not created JIRA tickets for it yet.

It fixes issues like JENKINS-34377 (restriction of PlaceHolder tasks by users in Job Restrictions) in a more generic way (see the current plugin-side fix in https://github.com/jenkinsci/job-restrictions-plugin/pull/20). With the such patch this plugin will start working as well as other plugins depending on `UpstreamCause#getUpstreamCauses()` resolution (e.g. https://github.com/search?p=1&q=org%3Ajenkinsci+getUpstreamCauses%28%29&type=Code)

@reviewbybees @msqb @jglick @abayer @svanoort

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -9,6 +9,9 @@ import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
 import hudson.console.ModelHyperlinkNote;
+import hudson.model.Actionable;
+import hudson.model.Cause;
+import hudson.model.CauseAction;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Item;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -92,8 +94,20 @@
...
     @Override
     public boolean start() throws Exception {
-        final PlaceholderTask task = new PlaceholderTask(getContext(), step.getLabel());
-        Queue.WaitingItem waitingItem = Queue.getInstance().schedule2(task, 0).getCreateItem();
+        final StepContext context = getContext();
+
+        // Construct cause to link the placeholder action
+        final Run<?, ?> run = context.get(Run.class);
+        final Cause.UpstreamCause upstreamCause;
+        if (run != null) {
```
jglick: No need to even check for null here.

oleg-nenashev:
@jglick @svanoort The story behind is https://github.com/jenkinsci/job-restrictions-plugin/pull/20 . I wanted to use a common `CauseAction` logic in order to get the standard upstream task resolution logic working for this case without patching in Job Restrictions. 

I agree I can do the same with `getOwnerTask()` though I need to ensure that `SubTask` is used properly everywhere. IIRC Matrix Project passes the cause though it is not the best example

jglick (changes requested):
Whatever this was doing, it is probably superseded by #170.

PR #59, opened 2017-12-15
[JENKINS-48575] notify RunListeners from PlaceHolderExecutable

ndeloof:
see https://github.com/jenkinsci/docker-plugin/blob/master/src/main/java/com/nirima/jenkins/plugins/docker/listener/DockerRunListener.java#L32 for concrete usage

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -20,6 +20,7 @@ import hudson.model.ResourceList;
 import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
+import hudson.model.listeners.RunListener;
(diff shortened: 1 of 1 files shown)


ndeloof:
This is unrelated to `dockerNode`. Goal here is to attach to a run an Action about the Computer(s) that has been used to run it. As a docker agent is transient, I want to collect information about the Agent as the build starts on it, so I can attach those runtime infos to the build for further diagnostic : typically `docker inspect` payload to track the actual physical node hosting the Agent. 

`ExecutorListener` only let us know about the task in queue, not about the (not yet created) Run.

I consider a "part of Pipeline X" PlaceHolderTask to be a legitimate a Run, considering it's definition in Jenkins API : this is (partial) execution of a (pipeline) Job on a Slave. PlaceHolderTask, even being a hack to get Slave allocated, is the only Pipeline component to match this API. So RunListener should be notified.

I don't know any other listener invoked to track execution of a Run on a Node. If you know a better way I'd be please to adopt it.

ndeloof:
inactive, feel free to take over

PR #60, opened 2018-01-12
Update index in log message to show the node name

dwnusbaum:
I noticed this while looking through the log of a failed build:

> after waiting for 300,000 ms because we assume unknown Node 300,000 is never going to appear!

@reviewbybees


dwnusbaum:
@reviewbybees done

Thanks!

PR #61, opened 2018-01-22
[JENKINS-46652] Simplifying SECURITY-675 test using core bug fix

jglick:
Downstream of https://github.com/jenkinsci/jenkins/pull/3254.

@reviewbybees


svanoort:
Eligible for merge when the core dependency bump is no longer pertinent (call it 6 months after release or when jenkins install stats show >50% are above that core version).

jglick:
I do not (much) care when it is merged. The point is to demonstrate the upstream fix. Evaluate it on that merit.

PR #63, opened 2018-02-09
[JENKINS-52165] Calling Controller.watch API

jglick:
[JENKINS-52165](https://issues.jenkins-ci.org/browse/JENKINS-52165) Split out of #21. Downstream of https://github.com/jenkinsci/durable-task-plugin/pull/60. Subsumes #73. Should improve performance by switching durable task logging and exit status checking from being pull (initiated by master at intervals) to push (initiated by agent at close intervals).

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
     <artifactId>workflow-durable-task-step</artifactId>
-    <version>2.18</version>
+    <version>2.19-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Nodes and Processes</name>
     <url>https://wiki.jenkins.io/display/JENKINS/Pipeline+Nodes+and+Processes+Plugin</url>
@@ -47,7 +47,7 @@
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-      <tag>workflow-durable-task-step-2.18</tag>
+      <tag>HEAD</tag>
   </scm>
     <repositories>
         <repository>
--- a/pom.xml
+++ b/pom.xml
@@ -75,7 +75,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.18-20180125.194359-1</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/57 -->
+            <version>1.18-20180209.181433-2</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.24-rc765.2c5d9ffed127</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/89 -->
+            <version>2.24</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.26-rc361.8feca397d0cb</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
+            <version>1.26-rc362.b077ce3ee7ca</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.26-rc362.b077ce3ee7ca</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
+            <version>1.26-rc366.c1ee6607c693</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.25-rc353.399e3eb7deb7</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
+            <version>1.26-rc361.8feca397d0cb</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/60 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -30,12 +30,14 @@ import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Main;
 import hudson.model.TaskListener;
 import hudson.util.DaemonThreadFactory;
 import hudson.util.FormValidation;
 import hudson.util.LogTaskListener;
 import hudson.util.NamingThreadFactory;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
@@ -46,8 +48,10 @@ import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.util.Timer;
+import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.durabletask.Controller;
 import org.jenkinsci.plugins.durabletask.DurableTask;
+import org.jenkinsci.plugins.durabletask.Handler;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.Step;
@@ -129,15 +133,20 @@ public abstract class DurableTaskStep extends Step {
 
     }
 
+    interface ExecutionRemotable {
+        void exited(int code, byte[] output) throws Exception;
+    }
+
     /**
      * Represents one task that is believed to still be running.
      */
     @SuppressFBWarnings(value="SE_TRANSIENT_FIELD_NOT_RESTORED", justification="recurrencePeriod is set in onResume, not deserialization")
-    static final class Execution extends AbstractStepExecutionImpl implements Runnable {
+    static final class Execution extends AbstractStepExecutionImpl implements Runnable, ExecutionRemotable {
 
         private static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
         private static final long MAX_RECURRENCE_PERIOD = 15000; // 15s
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
+        private static final long WATCHING_RECURRENCE_PERIOD = Main.isUnitTest ? /* 5s */5_000: /* 5m */300_000;
 
         private static final ScheduledThreadPoolExecutor THREAD_POOL = new ScheduledThreadPoolExecutor(25, new NamingThreadFactory(new DaemonThreadFactory(), DurableTaskStep.class.getName()));
         static {
@@ -156,6 +165,7 @@ public abstract class DurableTaskStep extends Step {
         private boolean returnStdout; // serialized default is false
         private String encoding; // serialized default is irrelevant
         private boolean returnStatus; // serialized default is false
+        private boolean watching;
 
         Execution(StepContext context, DurableTaskStep step) {
             super(context);
@@ -173,9 +183,16 @@ public abstract class DurableTaskStep extends Step {
             if (returnStdout) {
                 durableTask.captureOutput();
             }
-            controller = durableTask.launch(context.get(EnvVars.class), ws, context.get(Launcher.class), context.get(TaskListener.class));
+            TaskListener listener = context.get(TaskListener.class);
+            controller = durableTask.launch(context.get(EnvVars.class), ws, context.get(Launcher.class), listener);
             this.remote = ws.getRemote();
-            setupTimer();
+            try {
+                controller.watch(ws, new HandlerImpl(this, ws, listener), listener);
+                watching = true;
+            } catch (UnsupportedOperationException x) {
+                LOGGER.log(Level.WARNING, null, x);
+            }
+            setupTimer(watching ? WATCHING_RECURRENCE_PERIOD : MIN_RECURRENCE_PERIOD);
             return false;
         }
 
@@ -186,6 +203,19 @@ public abstract class DurableTaskStep extends Step {
                     LOGGER.log(Level.FINE, "Jenkins is not running, no such node {0}, or it is offline", node);
                     return null;
                 }
+                if (watching) {
+                    try {
+                        controller.watch(ws, new HandlerImpl(this, ws, listener()), listener());
+                        recurrencePeriod = WATCHING_RECURRENCE_PERIOD;
+                    } catch (UnsupportedOperationException x) {
+                        getContext().onFailure(x);
+                    } catch (Exception x) { // as below
+                        LOGGER.log(Level.FINE, node + " is evidently offline now", x);
+                        ws = null;
+                        recurrencePeriod = MIN_RECURRENCE_PERIOD;
+                        return null;
+                    }
+                }
             }
             boolean directory;
             try (Timeout timeout = Timeout.limit(10, TimeUnit.SECONDS)) {
@@ -194,6 +224,7 @@ public abstract class DurableTaskStep extends Step {
                 // RequestAbortedException, ChannelClosedException, EOFException, wrappers thereof; InterruptedException if it just takes too long.
                 LOGGER.log(Level.FINE, node + " is evidently offline now", x);
                 ws = null;
+                recurrencePeriod = MIN_RECURRENCE_PERIOD;
                 if (!printedCannotContactMessage) {
                     listener().getLogger().println("Cannot contact " + node + ": " + x);
                     printedCannotContactMessage = true;
@@ -203,6 +234,7 @@ public abstract class DurableTaskStep extends Step {
             if (!directory) {
                 throw new AbortException("missing workspace " + remote + " on " + node);
             }
+            LOGGER.log(Level.FINE, "{0} seems to be online", node);
             return ws;
         }
 
@@ -309,10 +341,20 @@ public abstract class DurableTaskStep extends Step {
                 return;
             }
             if (workspace == null) {
+                recurrencePeriod = Math.min((long) (recurrencePeriod * RECURRENCE_PERIOD_BACKOFF), MAX_RECURRENCE_PERIOD);
                 return; // slave not yet ready, wait for another day
             }
             TaskListener listener = listener();
             try (Timeout timeout = Timeout.limit(10, TimeUnit.SECONDS)) {
+                if (watching) {
+                    Integer exitCode = controller.exitStatus(workspace, launcher());
+                    if (exitCode == null) {
+                        LOGGER.log(Level.FINE, "still running in {0} on {1}", new Object[] {remote, node});
+                    } else {
+                        LOGGER.log(Level.FINE, "exited with {0} in {1} on {2}; expect asynchronous exit soon", new Object[] {exitCode, remote, node});
+                        // TODO if we get here again and exited has still not been called, assume we lost the notification somehow and end the step
+                    }
+                } else { // legacy mode
                         if (controller.writeLog(workspace, listener.getLogger())) {
                             getContext().saveState();
                             recurrencePeriod = MIN_RECURRENCE_PERIOD; // got output, maybe we will get more soon
@@ -337,6 +379,7 @@ public abstract class DurableTaskStep extends Step {
                             recurrencePeriod = 0;
                             controller.cleanup(workspace);
                         }
+                }
             } catch (Exception x) {
                 LOGGER.log(Level.FINE, "could not check " + workspace, x);
                 ws = null;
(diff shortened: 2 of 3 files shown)


jglick:
By the way https://github.com/jenkinsci/support-core-plugin/pull/128 should make it easier to see the impact of this on actual Remoting usage. Have not yet tried it myself.

oleg-nenashev:
Approved once the upstream PRs are integrated && the conflict is fixed

vivek:
This looks good to go if merge conflicts fixed? ping @jglick.

svanoort:
@vivek The upstream isn't really ready-to-go yet, and the whole thing needs some sort of proper and thorough testing in a simulated environment. Oleg's review found some issues already. 

We could of course *YOLO-merge* all this but I make no promises on the fallout if so.

jglick:
Well, the upstream “issues” are just things like Javadoc that should be improved, not user-visible problems, but yes this all deserves testing in a realistic environment. Once @svanoort is ready to start that I will polish up all the PRs.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -309,10 +352,20 @@ private void check() {
...
             }
             TaskListener listener = listener();
             try (Timeout timeout = Timeout.limit(10, TimeUnit.SECONDS)) {
+                if (watching) {
+                    Integer exitCode = controller.exitStatus(workspace, launcher());
+                    if (exitCode == null) {
+                        LOGGER.log(Level.FINE, "still running in {0} on {1}", new Object[] {remote, node});
+                    } else {
+                        LOGGER.log(Level.FINE, "exited with {0} in {1} on {2}; expect asynchronous exit soon", new Object[] {exitCode, remote, node});
+                        // TODO if we get here again and exited has still not been called, assume we lost the notification somehow and end the step
```
jglick: Reminding myself to check whether this concern is still valid.

pom.xml
```diff
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.24-rc765.2c5d9ffed127</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/89 -->
+            <version>2.24</version>
```
jglick: Really belongs in `master`; guess @svanoort forgot this during the release cycle.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +407,67 @@ private void check() {
...
+                    listener().getLogger().write(output); // diagnostic
+                }
+                getContext().onFailure(new AbortException("script returned exit code " + exitCode));
+            }
+        }
+
         @Override public void onResume() {
-            setupTimer();
+            ws = null; // find it from scratch please, rewatching as needed
+            setupTimer(MIN_RECURRENCE_PERIOD);
```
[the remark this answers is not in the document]
jglick: Never mind, I forgot that `getWorkspace` does both of these things when resumed in `watching` mode.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -347,17 +401,67 @@ private void check() {
...
+        private final ExecutionRemotable execution;
+        private final TaskListener listener;
+
+        HandlerImpl(Execution execution, FilePath workspace, TaskListener listener) {
+            this.execution = workspace.getChannel().export(ExecutionRemotable.class, execution);
+            this.listener = listener;
+        }
+
+        @Override public void output(InputStream stream) throws Exception {
+            IOUtils.copy(stream, listener.getLogger());
```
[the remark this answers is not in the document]
svanoort: > other network errors: coverage unknown

😒

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -158,6 +177,9 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
         private String remote;
         private boolean returnStdout; // serialized default is false
         private boolean returnStatus; // serialized default is false
+        private boolean watching; // serialized default is false
```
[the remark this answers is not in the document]
jglick: Right. For a `sh` step running across the upgrade, we behave in traditional polling mode.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -174,14 +196,23 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
...
-            controller = durableTask.launch(context.get(EnvVars.class), ws, context.get(Launcher.class), context.get(TaskListener.class));
+            controller = durableTask.launch(context.get(EnvVars.class), ws, context.get(Launcher.class), listener);
             this.remote = ws.getRemote();
-            setupTimer();
+            if (USE_WATCHING) {
+                try {
+                    controller.watch(ws, new HandlerImpl(this, ws, listener), listener);
+                    watching = true;
+                } catch (UnsupportedOperationException x) {
+                    LOGGER.log(Level.WARNING, null, x);
```
svanoort: Shouldn't we fail out here?

jglick: No, we just fall back to polling mode.

svanoort: Worth a code comment to mention that perhaps?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -252,6 +300,14 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
                             recurrencePeriod = 0;
                             listener().getLogger().println("After 10s process did not stop");
                             getContext().onFailure(cause);
+                            try {
+                                FilePath workspace = getWorkspace();
+                                if (workspace != null) {
+                                    controller.cleanup(workspace);
+                                }
+                            } catch (IOException | InterruptedException x) {
+                                Functions.printStackTrace(x, listener().getLogger());
```
svanoort: 🐜 Might be worth mentioning that we failed to do workspace cleanup (so it's clear this is not super serious).

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -315,10 +371,23 @@ private void check() {
...
                 return; // slave not yet ready, wait for another day
             }
             TaskListener listener = listener();
             try (Timeout timeout = Timeout.limit(10, TimeUnit.SECONDS)) {
+                if (watching) {
+                    Integer exitCode = controller.exitStatus(workspace, launcher());
+                    if (exitCode == null) {
+                        LOGGER.log(Level.FINE, "still running in {0} on {1}", new Object[] {remote, node});
+                    } else if (awaitingAsynchExit) {
+                        recurrencePeriod = 0;
```
svanoort: Maybe safer to have a *slight* nonzero delay?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
svanoort: This implementation means it's *absolutely* a requirement that the invoker know to re-send log content if it fails as the result of transitory network conditions.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +423,67 @@ private void check() {
             }
         }
 
+        // called remotely from HandlerImpl
+        @Override public void exited(int exitCode, byte[] output) throws Exception {
+            try {
+                getContext().get(TaskListener.class);
```
svanoort: Why are we checking for TaskListener in the context in order to identify this condition?

I'll take my answer in the form of a code comment, please.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
svanoort: @jglick If you're the author and couldn't remember that, it's probably worth a code comment so future maintainers (cough) don't make the same assumption and waste time.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: No, (exact) zero is the signal to avoid rescheduling.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
jglick: I do not think so. Network conditions are unlikely to be “transitory”. If there are some lost packets or whatever, TCP will recover and the application layer is not going to know any different.

If the socket is broken—or there is otherwise a serious error in the Remoting transport, say Kafka (under development)—we hope the `Channel` will get closed somehow, and we may subsequently reconnect and resume log transmission. In such cases, we are making a best effort to resume the log where we left off, but we cannot know exactly where that was—the source channel position is an estimate. At worst we lose, or duplicate, a bit of the log file, which is hardly a serious matter when a build is a low-value object (and there is a good chance this build would wind up failing anyway).

At any rate such considerations really only apply to on-master file storage; cloud storage will have completely different characteristics.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
jglick: I actually _did_ have a comment, but apparently need to expand it.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: ack

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Could do that.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: ack

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -192,26 +223,43 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
                     LOGGER.log(Level.FINE, "Jenkins is not running, no such node {0}, or it is offline", node);
                     return null;
                 }
+                if (watching) {
+                    try {
+                        controller.watch(ws, new HandlerImpl(this, ws, listener()), listener());
+                        recurrencePeriod = WATCHING_RECURRENCE_PERIOD;
+                    } catch (UnsupportedOperationException x) {
+                        getContext().onFailure(x);
```
dwnusbaum: Would it be useful to set `watching = false` and fall back to polling in this case? Is there even any way to get here without modifying the serialized state? I guess if the controller implementation supported `watch`, and then we changed the implementation and removed the override, then any pipelines that resume after upgrading this plugin would break, but that feels like a pretty contrived case. Maybe worth a comment that this should never happen IIUC.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
svanoort: > At any rate such considerations really only apply to on-master file storage; cloud storage will have completely different characteristics.

Actually, it's even worse in Cloud environments. 

I'll let Google make my argument here about how to design resilient distributed systems and exactly *how little* you can trust the network: http://www.hpcs.cs.tsukuba.ac.jp/~tatebe/lecture/h23/dsys/dsd-tutorial.html#Design

> Both clients and servers must be able to deal with unresponsive senders/receivers.

> When you design distributed systems, you have to say, "Failure happens all the time." So when you design, you design for failure. It is your number one concern. What does designing for failure mean? One classic problem is partial failure. If I send a message to you and then a network failure occurs, there are two possible outcomes. One is that the message got to you, and then the network broke, and I just didn't get the response. The other is the message never got to you because the network broke before it arrived.

I'll out of band cite a couple issues for you too.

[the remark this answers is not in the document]
jglick: > it's even worse in Cloud environments

It could be better or much worse, depending on the “cloud”, but at any rate it is irrelevant for purposes of this PR since the behavior is up the provider of the `TaskListener`.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Yes it should never happen; can make that clearer in a comment.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
svanoort: Hmm, this is actually worse than it seems, because apparently the `PrintStream` returned by `getLogger` doesn't even throw an exception on failed write operations performed by `IOUtils.copy(stream, listener.getLogger());` - it just sets a 'trouble' flag you have to fetch via 'checkError' :-/

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
svanoort: Ah, so it's a special signal value, normally that would be "-1".

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
svanoort: Thinking about this a bit deeper, I really think this is broken (a hard 🐛 ) but primarily because of the PrintStream having all-but nonexistent error handling.  There *must* be some way to signal and propagate errors from this process so implementations can signal "hey this thing is broken" and it can get bubbled up. 

I would have deep, deep reservations about treating the log as without value -- IMO half the value of Cloud storage for logs is that they can be audited, and if we don't make *some* effort to deal with failures, we've just lot all ability to audit.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Perhaps. At any rate, zero has been used for this purpose from the start—not introduced in this patch. I can make that explicit in Javadoc.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
jglick: > the value of Cloud storage for logs is that they can be audited, and if we don't make some effort to deal with failures

My only concern right now is tolerable behavior for the built-in Remoting (to master file) log sink in the face of corner cases. It is OK to lose a bit of text when the `Channel` is dying or dead: we likely already were in the polling implementation, and the build is often going to fail anyway with an error of no interest to developers. It is not OK to have the build hang indefinitely, a master `Thread` consumed, etc.—things which happened in earlier iterations of the polling implementation and are now I think defended against in most cases.

Cloudy storage of logs is a separate case. Quality of service levels are up to that implementation, not this generic code. I would envision that a production implementation would make some effort to buffer and resend messages in the background, giving up at some point. If it has some evidence that an outage is somehow tied to the state of the agent JVM and that an agent restart would actually solve something, then it could override `PrintStream.write` to throw `RuntimeException`, though I doubt this would be productive.

More of interest is a routine `ChannelClosedException` during master restart, for which we would prefer to have exact log content if at all possible. Not sure if that is effectively testable but can try it. One trick which may help is to extract the `FilterOutputStream.out`.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Cancel that—at this point we have already printed an unusual message above, and any exception here probably came from `getWorkspace`.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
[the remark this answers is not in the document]
jglick: > More of interest

(resolved)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -132,11 +142,25 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
...
+    @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "public & mutable only for tests")
+    @Restricted(NoExternalUse.class)
+    public static long WATCHING_RECURRENCE_PERIOD = /* 5m */300_000;
+
+    /** If set to false, disables {@link Execution#watching} mode. */
+    @SuppressWarnings("FieldMayBeFinal")
+    private static boolean USE_WATCHING = !"false".equals(System.getProperty(DurableTaskStep.class.getName() + ".USE_WATCHING"));
+
     /**
      * Represents one task that is believed to still be running.
```
jglick: I will flesh this out with an explanation of the expected state changes.

jglick: (done)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (done)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +484,111 @@ private void check() {
...
+            OutputStream os;
+            if (ps.getClass() == PrintStream.class) {
+                // Try to extract the underlying stream, since swallowing exceptions is undesirable and PrintStream.checkError is useless.
+                os = (OutputStream) printStreamDelegate.get(ps);
+            } else {
+                // A subclass. Who knows why, but trust any write(…) overrides it may have.
+                os = ps;
+            }
+            try {
+                IOUtils.copy(stream, os);
```
dwnusbaum: It looks like all `PrintStream#write` overloads synchronize on `this` and null-check `out` before writing (since it is nulled out once the stream is closed). Do we need to do that here as well to avoid breaking `PrintStream` invariants? The usage here should always be from the same thread IIUC, but do other threads have access to `listener.getLogger()`?

jglick: Agreed, that sounds safer.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -560,16 +560,17 @@ private void setupTimer(long initialRecurrencePeriod) {
...
-            } else {
-                // A subclass. Who knows why, but trust any write(…) overrides it may have.
-                os = ps;
-            }
             try {
-                IOUtils.copy(stream, os);
+                if (ps.getClass() == PrintStream.class) {
+                    // Try to extract the underlying stream, since swallowing exceptions is undesirable and PrintStream.checkError is useless.
+                    OutputStream os = (OutputStream) printStreamDelegate.get(ps);
+                    synchronized (os) { // like PrintStream.write overloads do
```
dwnusbaum: Did you mean `synchronized (ps)`?

jglick: Indeed that seems to be the lock, will change.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +484,111 @@ private void check() {
...
+            this.listener = listener;
+        }
+
+        @Override public void output(InputStream stream) throws Exception {
+            PrintStream ps = listener.getLogger();
+            try {
+                if (ps.getClass() == PrintStream.class) {
+                    // Try to extract the underlying stream, since swallowing exceptions is undesirable and PrintStream.checkError is useless.
+                    synchronized (ps) { // like PrintStream.write overloads do
+                        IOUtils.copy(stream, (OutputStream) printStreamDelegate.get(ps));
```
dwnusbaum: Is it possible for the PrintStream to be closed? If so, we'd get an NPE because `printStreamDelegate.get(ps)` will be null.

jglick: It _should_ not be closed. If it is, we expect this method to throw an exception. I can make it throw a specific exception for better diagnostics.

dwnusbaum: Ok, I guess it would be clear to anyone investigating where the NPE came from anyway, so it probably doesn't need a specific exception, but 3f8ced4 won't hurt.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +484,115 @@ private void check() {
...
         private static final long serialVersionUID = 1L;
 
     }
 
+    private static class HandlerImpl extends Handler {
+
+        private static final Field printStreamDelegate;
+        static {
+            try {
+                printStreamDelegate = FilterOutputStream.class.getDeclaredField("out");
```
svanoort: Ahh, reflection hacks: the universal panacea!

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -353,17 +484,115 @@ private void check() {
...
+                    // A subclass. Who knows why, but trust any write(…) overrides it may have.
+                    IOUtils.copy(stream, ps);
+                }
+            } catch (ChannelClosedException x) {
+                // We are giving up on this watch. Wait for some call to getWorkspace to rewatch.
+                throw x;
+            } catch (Exception x) {
+                // Try to report it to the master.
+                try {
+                    execution.problem(x);
```
svanoort: I predict we'll be seeing a lot of this particular report when there are network issues.   Thankfully it won't be *that* often due to the between recurrence periods.

PR #64, opened 2018-02-12
[JEP-206] Gather command output in a local encoding

jglick:
[JEP-206](https://github.com/jenkinsci/jep/blob/master/jep/206/README.adoc)

Extracted from #21. Downstream of https://github.com/jenkinsci/durable-task-plugin/pull/61. Also downstream of https://github.com/jenkinsci/workflow-job-plugin/pull/89 for testing, but really as a user you want to update `workflow-job` as well, unless your master happens to be running in UTF-8.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin()
+buildPlugin(jenkinsVersions: [null, '2.121.1'])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,1 +1,1 @@
-buildPlugin(jenkinsVersions: [null, '2.73.1'])
+buildPlugin()
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.14</version>
+        <version>3.19</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -67,7 +67,7 @@
         <jenkins.version>2.73.3</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
-        <workflow-support-plugin.version>2.19-rc287.918f5958551d</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/56 -->
+        <workflow-support-plugin.version>2.20-rc569.8b4964ac1954</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/56 -->
     </properties>
     <dependencies>
         <dependency>
@@ -78,7 +78,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.23-rc125.7208013b9494</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/61 -->
+            <version>1.24-rc319.e1ef73b3ab09</version> <!-- TODO pending release -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.22-rc329.5cb884ef714a</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/89 -->
+            <version>2.24-rc765.2c5d9ffed127</version> <!-- TODO https://github.com/jenkinsci/workflow-job-plugin/pull/89 -->
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <jenkins.version>2.73.3</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
-        <workflow-support-plugin.version>2.20-rc569.8b4964ac1954</workflow-support-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-support-plugin/pull/56 -->
+        <workflow-support-plugin.version>2.20</workflow-support-plugin.version>
     </properties>
     <dependencies>
         <dependency>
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -67,7 +69,7 @@
     private static final Logger LOGGER = Logger.getLogger(DurableTaskStep.class.getName());
 
     private boolean returnStdout;
-    private String encoding = DurableTaskStepDescriptor.defaultEncoding;
+    private String encoding;
```
jglick: Note that this is a behavioral change—we are changing the default value of `encoding` from `UTF-8` to null (i.e., use the node’s system encoding). Also the encoding now applies to regular streamed output, whereas previously it only mattered for `returnStdout: true`. (Which is the main point of this whole exercise: JENKINS-31096.)

I tried to think of a way to keep the previous default but it was just going to be ugly. With the change of the overall build log to UTF-8, it would mean that you would get mojibake in case you were running, say, entirely on CP-1252 Windows (with or without agents) and just did something simple like

```groovy
node {
  bat 'something producing Western European characters…'
}
```

which seems wrong.

This interpretation of `encoding` is also consistent with the `readFile` and `writeFile` steps, as well as the behavior of freestyle builds (which define the entire build’s log encoding by the node’s system encoding), and I think would be considered most intuitive going forward. Otherwise we would need a special value (`''`, `'system'`, …) indicating system default encoding, and then try to explain that in the snippet generator.

Obviously if you have specific knowledge that a given process is going to be producing UTF-8 yet will be running on a node with a different system encoding (typically Windows, since every Linux distribution has defaulted to UTF-8 for a long time), you can make that explicit with `encoding: 'UTF-8'`.

pom.xml
```diff
@@ -62,9 +62,10 @@
         </pluginRepository>
     </pluginRepositories>
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.73.3</jenkins.version>
```
svanoort: Is this gratuitous or do we need a new API?

jglick: Test dep on `workflow-job` requires this.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
svanoort: > Obviously if you have specific knowledge that a given process is going to be producing UTF-8 yet will be running on a node with a different system encoding (typically Windows, since every Linux distribution has defaulted to UTF-8 for a long time), you can make that explicit with encoding: 'UTF-8'.

🐛  This should be called out really explicitly in the documentation / UI here.

jglick: Yeah I can emphasize that.

jglick:
> is there a way we can execute these PRs so users don't have to update the plugins all in lockstep in order not to result in some mangled output?

No, as noted elsewhere you need to update both **Pipeline: Job** and **Pipeline: Nodes and Processes**. But if you care about Unicode, you probably have mangled output already, so :man_shrugging:

jglick:
and bye-bye @rsandell

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
oleg-nenashev: Seems reasonable. Whatever is done with encodings in Durable Task and this plugin, it is a risky change anyway. Since the idea is to NOT enforce the UTF encoding, the current approach looks preferable

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -102,17 +104,15 @@ public boolean isReturnStatus() {
 
     public abstract static class DurableTaskStepDescriptor extends StepDescriptor {
 
-        public static final String defaultEncoding = "UTF-8";
-
         public FormValidation doCheckEncoding(@QueryParameter boolean returnStdout, @QueryParameter String encoding) {
+            if (encoding.isEmpty()) {
```
oleg-nenashev: `StringUtils.isBlank()`? otherwise the behavior will differ from `Util.fixEmpty`
🐜

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -173,6 +171,11 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
             if (returnStdout) {
                 durableTask.captureOutput();
             }
+            if (step.encoding != null) {
+                durableTask.charset(Charset.forName(step.encoding));
```
oleg-nenashev: Catch/rethrow exception with more debug details?

Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.73.1'])
+buildPlugin()
```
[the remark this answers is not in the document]
jglick: Agreed.

pom.xml
```diff
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.2</version>
+        <version>3.4</version>
```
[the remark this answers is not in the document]
jglick: Now 3.7.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Which debug details are you looking for here?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: This is a web method; the text field may be blank but not missing.

jglick:
Test failure looks like a flake.

carlossg:
@svanoort can you check the comments addressing your review ?

svanoort:
@carlossg I'm mostly booked up for the week due to the Java 10 hackathon, but will take a look as soon as we're out of that, thanks.

svanoort:
@jglick Although I can't tell if that test failure is a flake or actual issue here?

jglick:
> I can't tell if that test failure is a flake or actual issue

IIRC there were subsuming PRs which did _not_ fail, strongly suggesting a flake. I can try to retrigger the build.

PR #65, opened 2018-02-13
Watching plus Unicode

jglick:
Integrates #63 & #64. Downstream of https://github.com/jenkinsci/durable-task-plugin/pull/62.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -30,12 +30,14 @@ import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Main;
 import hudson.model.TaskListener;
 import hudson.util.DaemonThreadFactory;
 import hudson.util.FormValidation;
 import hudson.util.LogTaskListener;
 import hudson.util.NamingThreadFactory;
 import java.io.IOException;
+import java.io.InputStream;
 import java.nio.charset.Charset;
 import java.util.Set;
 import java.util.concurrent.ScheduledFuture;
@@ -46,8 +48,10 @@ import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.util.Timer;
+import org.apache.commons.io.IOUtils;
 import org.jenkinsci.plugins.durabletask.Controller;
 import org.jenkinsci.plugins.durabletask.DurableTask;
+import org.jenkinsci.plugins.durabletask.Handler;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.Step;
@@ -129,15 +133,20 @@ public abstract class DurableTaskStep extends Step {
 
     }
 
+    interface ExecutionRemotable {
+        void exited(int code, byte[] output) throws Exception;
+    }
+
     /**
      * Represents one task that is believed to still be running.
      */
     @SuppressFBWarnings(value="SE_TRANSIENT_FIELD_NOT_RESTORED", justification="recurrencePeriod is set in onResume, not deserialization")
-    static final class Execution extends AbstractStepExecutionImpl implements Runnable {
+    static final class Execution extends AbstractStepExecutionImpl implements Runnable, ExecutionRemotable {
 
         private static final long MIN_RECURRENCE_PERIOD = 250; // ¼s
         private static final long MAX_RECURRENCE_PERIOD = 15000; // 15s
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
+        private static final long WATCHING_RECURRENCE_PERIOD = Main.isUnitTest ? /* 5s */5_000: /* 5m */300_000;
 
         private static final ScheduledThreadPoolExecutor THREAD_POOL = new ScheduledThreadPoolExecutor(25, new NamingThreadFactory(new DaemonThreadFactory(), DurableTaskStep.class.getName()));
         static {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -102,17 +108,15 @@ public boolean isReturnStatus() {
 
     public abstract static class DurableTaskStepDescriptor extends StepDescriptor {
 
-        public static final String defaultEncoding = "UTF-8";
-
         public FormValidation doCheckEncoding(@QueryParameter boolean returnStdout, @QueryParameter String encoding) {
```
oleg-nenashev: Restrict it?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -102,17 +108,15 @@ public boolean isReturnStatus() {
...
+            if (encoding.isEmpty()) {
```
oleg-nenashev: Check for null just in case the form is messed up. Ideally it makes sense to use `StringUtils.isBlank()` here and in the constructor

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -186,6 +206,19 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
...
                     return null;
                 }
+                if (watching) {
+                    try {
+                        controller.watch(ws, new HandlerImpl(this, ws, listener()), listener());
+                        recurrencePeriod = WATCHING_RECURRENCE_PERIOD;
+                    } catch (UnsupportedOperationException x) {
+                        getContext().onFailure(x);
+                    } catch (Exception x) { // as below
+                        LOGGER.log(Level.FINE, node + " is evidently offline now", x);
```
oleg-nenashev: Apparently it's not documented in https://github.com/jenkinsci/durable-task-plugin/pull/60

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -203,6 +237,7 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
             if (!directory) {
                 throw new AbortException("missing workspace " + remote + " on " + node);
             }
+            LOGGER.log(Level.FINE, "{0} seems to be online", node);
```
oleg-nenashev: DEBUG/FINEST? maybe also makes sense to print Execution#toString() to the log so we understand from where it comes

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
oleg-nenashev: Execution link in the log would be helpful

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -85,7 +91,7 @@ public String getEncoding() {
     }
 
     @DataBoundSetter public void setEncoding(String encoding) {
-        this.encoding = encoding;
+        this.encoding = Util.fixEmpty(encoding);
```
oleg-nenashev: Does not handle blank strings 🐜

Jenkinsfile
```diff
@@ -1 +1 @@
-buildPlugin(jenkinsVersions: [null, '2.73.1'])
```
oleg-nenashev: Test with 2.107 instead?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Sorry, what?

> Convert empty string to null.

That is exactly what I want here.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Not part of this PR, but sure.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: If the form is messed up, there is a product bug and validation will fail. No need to be unreasonably defensive.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Sure.

jglick:
Please do _not_ perform line-by-line reviews in this PR. Rather, review component PRs.

Jenkinsfile
jglick: (obsolete)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (anyway for #64)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (for #64)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (for #64)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (for #63)

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: (#63)

jglick:
Now covered by #63.

PR #66, opened 2018-03-08
[DO-NOT-MERGE] Bump durable task to check for Windows behaviors in CI

svanoort:
Workaround for not having a handy Windows test environment.   Checking for JENKINS-34150 regression here to see if unit tests catch it.


jglick:
Apparently they did not, so is there a reason to still keep this open?

PR #67, opened 2018-03-22
Fixing an error in ExecutorStepTest.authentication

jglick:
A little rabbit hole here. I had jotted down but temporarily set aside a (nonfatal) stack trace printed during a recently added test, `ExecutorStepTest.authentication`:

```
java.io.IOException: cannot find current thread
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.doGet(CpsStepContext.java:295)
	at org.jenkinsci.plugins.workflow.support.DefaultStepContext.get(DefaultStepContext.java:61)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$1.run(ExecutorStepExecution.java:112)
	at …
```

Now first of all, it is a little extreme to print a system warning just because we are unable to get a reference to a build log to print an informational message to it. In the test

```groovy
timeout(time: 5, unit: 'SECONDS') {node {error 'should not be allowed'}}
```

the message is printed 15s after the `node` step starts, by which time the build has already aborted—fine, we do not need to print anything to the log of a completed build!

But then I thought, wait, why is that `Queue.Item` still there? After all, `stop` had already been called by the `timeout` step, and this is supposed to cancel any outstanding queue items associated with the `node` block. Turns out that in this case the `dev` user lacked enough permission to `Item.READ` that job, so `Queue.items` was coming back as an empty array. (A little artificial, since the build is running “as” this user, so normally they would be able to read it too, but whatever.) Really `Queue.getItems` needs to run as `SYSTEM` to make sure all items are checked—we are already looking for a specific item so this is not an information leak. OK, with that fixed, the item is successfully cancelled, but…

…then you get a nasty error from `CpsStepExecution` because it is getting aborted _twice_: once by `super.stop` (from `timeout`); once by `CancelledItemListener`, because now the placeholder task is getting cancelled. And trying to stop the same step twice with different causes is currently treated as a warning (code bug somewhere). Anyway the `AbortException` was delivered first, making the build a `FAILURE`, not `ABORTED`. So fixing the logic so that only a single `onFailure` is delivered, with the user abort.

Got all that? @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -1,6 +1,5 @@
 package org.jenkinsci.plugins.workflow.support.steps;
 
-import com.google.inject.Inject;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.EnvVars;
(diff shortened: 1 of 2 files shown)


svanoort:
Feel free to merge/release -- 3 approvals is definitely enough, and Devin has touched this code recently :)

dwnusbaum:
I will merge/release tomorrow.

PR #68, opened 2018-05-08
[JENKINS-54356] Add to node API for pipeline builds.

roguishmountain:
Based off of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/19.

Implements [JENKINS-54356](https://issues.jenkins-ci.org/browse/JENKINS-54356)

The node API currently shows run details for freestyle builds, but not pipeline builds.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -141,12 +141,5 @@
             <artifactId>structs</artifactId>
             <version>1.7</version>
         </dependency>
-        <dependency>
-            <groupId>com.googlecode.json-simple</groupId>
-            <artifactId>json-simple</artifactId>
-            <version>1.1.1</version>
-            <scope>test</scope>
-            <type>jar</type>
-        </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -141,5 +141,12 @@
             <artifactId>structs</artifactId>
             <version>1.7</version>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
+            <version>1.1.1</version>
+            <scope>test</scope>
+            <type>jar</type>
+        </dependency>
     </dependencies>
 </project>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -8,6 +8,7 @@ import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Util;
 import hudson.console.ModelHyperlinkNote;
+import hudson.model.Api;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Item;
@@ -663,7 +664,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
          * Occupies {@link Executor} while workflow uses this slave.
          */
         @ExportedBean
-        private final class PlaceholderExecutable implements ContinuableExecutable {
+        private final class PlaceholderExecutable implements ContinuableExecutable, AccessControlled {
 
             @Override public void run() {
                 final TaskListener listener;
@@ -786,7 +787,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             @Exported
             public Integer getNumber() {
                 Run<?, ?> r = getParent().runForDisplay();
-                return r != null ? r.getNumber() : -1;
+                return r != null ? r.getNumber() : null;
             }
 
             @Exported
@@ -845,6 +846,38 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             }
 
             private static final long serialVersionUID = 1L;
+
+            @Override
+            public ACL getACL() {
+                try {
+                    if (!context.isReady()) {
+                        return Jenkins.getActiveInstance().getACL();
+                    }
+                    FlowExecution exec = context.get(FlowExecution.class);
+                    if (exec == null) {
+                        return Jenkins.getActiveInstance().getACL();
+                    }
+                    Queue.Executable executable = exec.getOwner().getExecutable();
+                    if (executable instanceof AccessControlled) {
+                        return ((AccessControlled) executable).getACL();
+                    } else {
+                        return Jenkins.getActiveInstance().getACL();
+                    }
+                } catch (Exception x) {
+                    LOGGER.log(FINE, null, x);
+                    return Jenkins.getActiveInstance().getACL();
+                }
+            }
+
+            @Override
+            public void checkPermission(Permission permission) throws AccessDeniedException {
+                getACL().checkPermission(permission);
+            }
+
+            @Override
+            public boolean hasPermission(Permission permission) {
+                return getACL().hasPermission(permission);
+            }
         }
 
         private static final long serialVersionUID = 1098885580375315588L; // as of 2.12
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -849,24 +849,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
 
             @Override
             public ACL getACL() {
-                try {
-                    if (!context.isReady()) {
-                        return Jenkins.getActiveInstance().getACL();
-                    }
-                    FlowExecution exec = context.get(FlowExecution.class);
-                    if (exec == null) {
-                        return Jenkins.getActiveInstance().getACL();
-                    }
-                    Queue.Executable executable = exec.getOwner().getExecutable();
-                    if (executable instanceof AccessControlled) {
-                        return ((AccessControlled) executable).getACL();
-                    } else {
-                        return Jenkins.getActiveInstance().getACL();
-                    }
-                } catch (Exception x) {
-                    LOGGER.log(FINE, null, x);
-                    return Jenkins.getActiveInstance().getACL();
-                }
+                return getParent().getACL();
             }
 
             @Override
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -634,7 +634,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
          * Occupies {@link Executor} while workflow uses this slave.
          */
         @ExportedBean
-        public final class PlaceholderExecutable implements ContinuableExecutable {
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -141,5 +141,12 @@
             <artifactId>structs</artifactId>
             <version>1.7</version>
         </dependency>
+        <dependency>
+            <groupId>com.googlecode.json-simple</groupId>
+            <artifactId>json-simple</artifactId>
```
dwnusbaum: Jenkins core already contains a library for working with JSON: https://github.com/kohsuke/Json-lib (Classes are in the net.sf.json package (e.g. net.sf.json.JSONObject)), so I would prefer to reuse that library instead of adding a new dependency.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -450,6 +453,45 @@ void assertLogMatches(WorkflowRun build, String regexp) throws IOException { //
...
     }
 
+    @Test public void detailsExported() throws Exception {
+        story.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                DumbSlave s = story.j.createOnlineSlave();
+
+                JSONParser parser = new JSONParser();
+                JSONObject propertiesJSON = null;
```
dwnusbaum: nit: No need to declare and initialize the variable to null up here, you can just declare and assign it directly below where it is used.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -633,7 +634,7 @@ public void run() {
          * Occupies {@link Executor} while workflow uses this slave.
          */
         @ExportedBean
-        private final class PlaceholderExecutable implements ContinuableExecutable {
+        public final class PlaceholderExecutable implements ContinuableExecutable {
```
dwnusbaum: Why do we have to make the class public?

roguishmountain: I thought it needed to be public to use `@ExportedBean`, but that seems to be incorrect

dwnusbaum: Thanks for checking! I wasn't sure either way but figured that since it had `@ExportedBean` when it was private then maybe it didn't need to be public. If you are curious about the details, the exporting code appears to be in [Stapler's org.kohsuke.stapler.export package](https://github.com/stapler/stapler/tree/cafcb607e9e957bc8f57782eab3b0568a2ab3f42/core/src/main/java/org/kohsuke/stapler/export), although I have never looked at the code in detail.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -782,10 +783,33 @@ public void run() {
                 return PlaceholderTask.this;
             }
 
+            @Exported
+            public Integer getNumber() {
+                Run<?, ?> r = getParent().runForDisplay();
+                return r != null ? r.getNumber() : -1;
```
dwnusbaum: I think it was fine (probably preferable) to return `null` here as you had it previously - it just looked like the test needed to be fixed to handle `null` correctly.

roguishmountain:
after merging the changes from master `currentExecutable` is null during the `detailsExported()` test

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
roguishmountain: I'm also not sure. I thought you knew. I've only seen `@ExportedBean` with public classes, but that doesn't mean its correct

roguishmountain:
This was working locally, until I merged in the changes from master. `currentExecutable` is null during the `detailsExported()` test. After running `s.getComputer().getAllExecutors()` the value for `getCurrentExecutable()` was the excutable, but `getCurrentExecutableForApi()` was null. 

In Executor.java `getCurrentExecutableForApi()` calls `getCurrentExecutable()` and then returns `candidate instanceof AccessControlled && ((AccessControlled) candidate).hasPermission(Item.READ) ? candidate : null`. The issue is that for WorkflowJobs  `candidate instanceof AccessControlled` is false. 

I think this is due to this commit https://github.com/jenkinsci/jenkins/commit/fc4f6affd2a05864ba0cc7ae92da6bc04c8a35d9#diff-4da8c72b731d029f61f01ed41568d209 where the api is switched from exporting `getCurrentExecutable()` to `getCurrentExecutableForApi()`. Looking at https://javadoc.jenkins.io/hudson/security/AccessControlled.html under "All Known Implementing Classes" I see freestyle builds, but not workflow. I think this is what is causing the test to fail.

dwnusbaum:
@roguishmountain That list of "All Known Implementing Classes" only includes classes in Jenkins core, and `WorkflowJob` extends `Job`, which is in that list, and `WorkflowRun` extends `Run`, which is also in the list, and you don't set up any security on Jenkins for your test, so those aren't the problem, but you were on the right track. In our case, the `Queue.Executable` owned by the `Executor` is an instance of `ExecutorStepExecuton.PlaceholderTask.PlaceholderExecutable`, which does not implement `AccessControlled`. The following patch fixed things on my machine (I also changed `getNumber` back to returning null):

    diff --git a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
    index 7a56117..b98ffc3 100644
    --- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
    +++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
    @@ -664,5 +664,5 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
              */
             @ExportedBean
    -        private final class PlaceholderExecutable implements ContinuableExecutable {
    +        private final class PlaceholderExecutable implements ContinuableExecutable, AccessControlled {
     
                 @Override public void run() {
    @@ -787,5 +787,5 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 public Integer getNumber() {
                     Run<?, ?> r = getParent().runForDisplay();
    -                return r != null ? r.getNumber() : -1;
    +                return r != null ? r.getNumber() : null;
                 }
     
    @@ -846,4 +846,20 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
     
    +
    +            @Override
    +            public ACL getACL() {
    +                Run<?,?> r = runForDisplay();
    +                return r != null ? r.getACL() : Jenkins.getActiveInstance().getACL();
    +            }
    +
    +            @Override
    +            public void checkPermission(Permission prmsn) throws AccessDeniedException {
    +                getACL().checkPermission(prmsn);
    +            }
    +
    +            @Override
    +            public boolean hasPermission(Permission prmsn) {
    +                return getACL().hasPermission(prmsn);
    +            }
             }
     
I'm not sure if my `getACL` implementation is totally correct (maybe check to see what `ExecutorStepExecution.PlaceholderTask` does and copy that?), but it should be on the right track.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -797,16 +822,62 @@ public void run() {
...
+
             @Override public String toString() {
                 return "PlaceholderExecutable:" + PlaceholderTask.this;
             }
 
             private static final long serialVersionUID = 1L;
+
+            @Override
+            public ACL getACL() {
+                try {
```
dwnusbaum: I think it would be better to just use `return getParent().getACL();` since the goal is to use identical logic.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -527,6 +532,42 @@ void assertLogMatches(WorkflowRun build, String regexp) throws IOException { //
...
+                        + "semaphore 'wait'\n"
+                        + "    sleep 10\n"
+                        + "}"));
+
+                WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+                SemaphoreStep.waitForStart("wait/1", b);
+                JenkinsRule.WebClient wc = story.j.createWebClient();
+                Page page = wc
+                        .goTo("computer/" + s.getNodeName()
+                                + "/api/json?tree=executors[currentExecutable[number,displayName,url,timestamp]]", "application/json");
```
dwnusbaum: nit: Would be good to also test `fullDisplayName`.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -527,6 +532,42 @@ void assertLogMatches(WorkflowRun build, String regexp) throws IOException { //
...
+
+                JSONObject propertiesJSON = (JSONObject) (new JsonSlurper()).parseText(page.getWebResponse().getContentAsString());
+                JSONArray executors = propertiesJSON.getJSONArray("executors");
+                JSONObject executor = executors.getJSONObject(0);
+                JSONObject currentExecutable = executor.getJSONObject("currentExecutable");
+
+                assertEquals(1, currentExecutable.get("number"));
+
+                assertEquals("part of " + p.getName() + " #1",
```
dwnusbaum: nit: It looks like this message is actually [constructed using `WorkflowRun#getFullDisplayName`](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/5224a7b917127a66d6a14980f1ff98c614af2a6d/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L471), so it would be better to use `"part of " + b.getFullDisplayName()`, although it shouldn't actually change the text in this case.

roguishmountain: should getName and also getFullDisplayName be tested? or just getFullDisplayName?

dwnusbaum: Did you mean to respond to https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/68/files#r214370377? My thought there was just that it would be good to test all of the methods that you are adding `@Exported` to. You already test most of them: `tree=executors[currentExecutable[number,displayName,url,timestamp]`, but it would be good to add `fullDisplayName` to the tree parameter as well.

My point here is just that the text returned by `currentExecutable.get("displayName"))` would not match `"part of " + p.getName() + " #1"` if you changed the display name of the WorkflowRun, or placed the WorkflowJob in a folder, or if it wasn't the first build, etc., so using `"part of " + b.getFullDisplayName()` instead makes the test slightly more robust.

PR #69, opened 2018-05-15
Mostly sync ExecutorStep/help-label.html with AbstractProject/help-label.html

jsoref:
There were a couple of inconsistencies that were problematic...

Most notably `|` instead of `||`, but also `expr||expr`, and generally the lack of spacing around tokens, and the loss of other markup.

Afaict, the same display engine is used for both places, so I don't see any particular reason not to synchronize these two files.

Note that I'm moving the comment about the source file outside of the `<div>`, depending on the display engine, that could break everything...

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep/help-label.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep/help-label.html
@@ -1,5 +1,5 @@
 <div>
-    Computer name, label name, or any other label expression like <code>linux && 64bit</code> to restrict where this step builds.
+    Computer name, label name, or any other label expression like <code>linux &amp;&amp; 64bit</code> to restrict where this step builds.
     May be left blank, in which case any available executor is taken.
     <!-- TODO figure out how to inline or otherwise share AbstractProject/help-label.html -->
     <h3>Valid Operators</h3>
--- a/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep/help-label.html
+++ b/src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep/help-label.html
@@ -24,7 +24,7 @@
 
         <dt>a -> b</dt>
         <dd>
-            "implies" operator. Equivalent to <tt>!a|b</tt>.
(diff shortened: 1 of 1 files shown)


src/main/resources/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep/help-label.html
```diff
@@ -1,46 +1,107 @@
...
+      OR — <b>either</b> of the expressions <tt>a</tt> or <tt>b</tt> may be
+      true
+    </dd>
+
+    <dt><tt>a -> b</tt></dt>
+    <dd>
+      "implies" operator — equivalent to <tt>!a || b</tt>.<br>
+      For example, <tt>windows -> x64</tt> could be thought of as "if a Windows
+      agent is used, then that agent <i>must</i> be 64-bit", while still
+      allowing this project to be executed on any agents that <i>do not</i> have
```
dwnusbaum: It seems better to me to replace `project` with `step` throughout the file, since a Pipeline job (what I think of when I read `project`) can have multiple `node` steps, so using `project` is a little confusing.

jsoref: Actually, I think you mean `stage`

dwnusbaum: `stage` has a different meaning, right? Maybe `block` is the best way to refer to what runs inside of the `node` `step`? CC @abayer

jsoref: And, then, that gets a bit confusing, because sometimes it's the pipeline and sometimes it's the stage.

I think the right word ends up being `context`

abayer: `block` is probably the best option - in Declarative land, yes, behind the scenes we run a `node` step for the whole Pipeline and/or for individual `stage`s, but the `node` step isn't invoked by the user directly, so the docs for the step aren't really relevant there.

dwnusbaum:
Thanks @jsoref!

PR #71, opened 2018-06-07
Incrementalified

jglick:


svanoort:
Pending build of course

jglick:
Build passes.

PR #75, opened 2018-09-20
[JENKINS-28822] Distinguish between abort and failure for durable tasks

dwnusbaum:
See [JENKINS-28822](https://issues.jenkins-ci.org/browse/JENKINS-28822).

This PR uses [Jesse's proposed fix](https://issues.jenkins-ci.org/browse/JENKINS-28822?focusedCommentId=305068&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-305068) of storing any `Throwable` passed to `Execution#stop` so that it can be passed to `StepContext#onFailure` in `Execution#exited` instead of creating a new `AbortException`.

I am concerned about causing regressions for anyone relying on the current behavior or for anyone using the workarounds posted in JENKINS-28822. I will test the latter, but any thoughts on the former are welcome.

- [x] Confirm that the tests pass on Windows 
- [x] Test how the behavior of the workarounds posted in JENKINS-28822 are affected by this PR (EDIT: 1 workaround doesn't appear to work in the first place, and the other is unaffected by this PR for manual aborts, and affected in the same way as not using a workaround at all for timeout-induced aborts)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -239,6 +239,8 @@ public abstract class DurableTaskStep extends Step {
         private boolean watching; // serialized default is false
         /** Only used when {@link #watching}, if after {@link #WATCHING_RECURRENCE_PERIOD} comes around twice {@link #exited} has yet to be called. */
         private transient boolean awaitingAsynchExit;
+        /** The first throwable used to stop the task */
+        private transient volatile Throwable causeOfStoppage;
 
         Execution(StepContext context, DurableTaskStep step) {
             super(context);
@@ -350,6 +352,7 @@ public abstract class DurableTaskStep extends Step {
         }
 
         @Override public void stop(final Throwable cause) throws Exception {
+            causeOfStoppage = cause;
             FilePath workspace = getWorkspace();
             if (workspace != null) {
                 listener().getLogger().println("Sending interrupt signal to process");
@@ -503,13 +506,20 @@ public abstract class DurableTaskStep extends Step {
                 return;
             }
             recurrencePeriod = 0;
-            if (returnStatus || exitCode == 0) {
+            Throwable originalCause = causeOfStoppage;
+            if ((returnStatus && originalCause == null) || exitCode == 0) {
                 getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(output, StandardCharsets.UTF_8) : null);
             } else {
                 if (returnStdout) {
                     listener().getLogger().write(output); // diagnostic
                 }
-                getContext().onFailure(new AbortException("script returned exit code " + exitCode));
+                if (originalCause != null) {
+                    // JENKINS-28822: Use the previous cause instead of throwing a new AbortException
+                    listener().getLogger().println("script returned exit code " + exitCode);
+                    getContext().onFailure(originalCause);
+                } else {
+                    getContext().onFailure(new AbortException("script returned exit code " + exitCode));
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -239,6 +239,8 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
         private boolean watching; // serialized default is false
         /** Only used when {@link #watching}, if after {@link #WATCHING_RECURRENCE_PERIOD} comes around twice {@link #exited} has yet to be called. */
         private transient boolean awaitingAsynchExit;
+        /** The first throwable used to stop the task */
+        private transient volatile Throwable causeOfStoppage;
```
dwnusbaum: I think this needs to be `volatile` for access from both `#stop` and `#exited`, but if so then I think `recurrencePeriod` needs to be volatile as well (`long`s might be effectively volatile on 64-bit machines, but that is not guaranteed by the [JLS](https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.7)).

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -288,6 +288,40 @@ public DescriptorImpl() {
...
+                "  timeout(time: 1, unit: 'SECONDS') {" +
+                (Functions.isWindows()
+                        ? "bat 'ping -n 6 127.0.0.1 >nul'\n"
+                        : "sh 'sleep 5'\n") +
+                "  }" +
+                "}", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        j.waitForCompletion(b);
+        // Fails with Result.FAILURE as of commit a224295d12.
+        j.assertBuildStatus(Result.ABORTED, b);
```
dwnusbaum: I could also add some assertions that show what exceptions are thrown if reviewers think that would be useful.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: `stop` and `exited` could be called from unknown threads, so `volatile` seems appropriate.

`transient` is fine I think. There is a 10s grace period before forcible step termination, so that is the maximum we are dealing with. At worst, if there happens to be a restart at that moment, it would merely make this fix not take effect, thus yielding a `FAILURE`.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -288,6 +288,40 @@ public DescriptorImpl() {
...
+        p.setDefinition(new CpsFlowDefinition("node {\n" +
+                "  timeout(time: 1, unit: 'SECONDS') {" +
+                (Functions.isWindows()
+                        ? "bat 'ping -n 6 127.0.0.1 >nul'\n"
+                        : "sh 'sleep 5'\n") +
+                "  }" +
+                "}", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        j.waitForCompletion(b);
+        // Fails with Result.FAILURE as of commit a224295d12.
```
kutzi: ?
Is this a C&P'ed comment?

dwnusbaum: No, but it looks like I used a commit hash from durable-task (https://github.com/jenkinsci/durable-task-plugin/commit/a224295d12), which is wrong. I'll add a PR to fix it, thanks for noticing!

dwnusbaum: See #83.

PR #77, opened 2018-09-28
[JENKINS-53837] Associate node steps with their WorkflowJob correctly after restarts

dwnusbaum:
See [JENKINS-53837](https://issues.jenkins-ci.org/browse/JENKINS-53837).

This fixes an issue with custom `QueueTaskDispatcher`s trying to use `getOwnerTask` as part of their dispatching logic for restarted Pipelines.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -911,14 +911,10 @@ public class ExecutorStepTest {
         story.then(r -> {
             WorkflowJob p = r.jenkins.getItemByFullName("p1", WorkflowJob.class);
             WorkflowRun b = p.getBuildByNumber(1);
-            SemaphoreStep.waitForStart("wait/1", b);
             SemaphoreStep.success("wait/1", null);
-            while (b.isBuilding()) {
-                // This assertion fails before the fix for JENKINS-53837.
-                r.assertLogNotContains("Non-Pipeline tasks are forbidden!", b);
-                Thread.sleep(100);
-            }
+            r.waitForCompletion(b);
             r.assertBuildStatusSuccess(b);
+            r.assertLogNotContains("Non-Pipeline tasks are forbidden!", b);
         });
     }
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -916,4 +948,28 @@ String n(Run<?, ?> b, String label) {
         }
     }
 
+    @TestExtension("queueTaskOwnerCorrectWhenRestarting")
+    public static class PipelineOnlyTaskDispatcher extends QueueTaskDispatcher {
+        @Override
+        public CauseOfBlockage canTake(Node node, Queue.BuildableItem item) {
+            final Queue.Task[] t = new Queue.Task[]{item.task};
```
jglick: Clearer to either use a `final AtomicReference<Queue.Task>`, or just a (non`final`) `Queue.Task` and assign to a `final` variable before creating the anonymous inner class.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -890,6 +894,34 @@ String n(Run<?, ?> b, String label) {
...
+            p.setDefinition(new CpsFlowDefinition("node {\n" +
+                    "  semaphore('wait')\n" +
+                    "}", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", b);
+        });
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p1", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            SemaphoreStep.waitForStart("wait/1", b);
```
jglick: Superfluous, can be removed.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -890,6 +894,34 @@ String n(Run<?, ?> b, String label) {
...
+            SemaphoreStep.success("wait/1", null);
+            while (b.isBuilding()) {
```
jglick: A bit convoluted. I would think it would suffice to say something like

```java
SemaphoreStep.success("wait/1", null);
r.waitForCompletion(b);
r.assertBuildStatusSuccess(b);
r.assertLogNotContains("Non-Pipeline tasks are forbidden!", b);
```

dwnusbaum: Yeah, this was helpful for testing so that the test would fail immediately rather than timing out waiting, but I will clean it up now.

PR #78, opened 2018-10-01
[JENKINS-36547] Allow node reuse for pipeline jobs

orgi:
JENKINS-36547: Implement the getAffinityKey() method in order to make pipeline jobs try to reuse the node which they have been used last time.
Added couple of tests. They take quite some time. Please let me know if you think that I need to speed up and/or reduce testing.

Related PR: jenkinsci/jenkins#3611


orgi:
Unfortunately I cannot get this PR running on Jenkins. Seems like it would always try to test it against Jenkins version 2.121.1 in addition to the specified minimal required version (from incrementals). Of course using the old version it can not work. How to fix that?

dwnusbaum:
@orgi Because you don't have write permission to the repository, your changes to the Jenkinsfile are ignored. I will refile the PR myself so that you can get an accurate CI build to show that the upstream PR works.

dwnusbaum:
@orgi See #79

orgi:
Great, thanks :)
So I assume currently there's no task left open for me, isn't it?

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.23</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.73.3</jenkins.version>
+        <jenkins.version>2.145-rc27341.296ad9122b93</jenkins.version>
```
jglick: As a courtesy please indicate the upstream PR in a comment:

```xml
<jenkins.version>2.145-rc27341.296ad9122b93</jenkins.version> <!-- TODO https://github.com/jenkinsci/jenkins/pull/3611 -->
```

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -688,6 +688,328 @@ public boolean apply(@Nullable FlowNode input) {
...
+                }
+
+                WorkflowJob p = story.j.jenkins.createProject(WorkflowJob.class, "demo");
+                p.setDefinition(new CpsFlowDefinition("node {echo \"ran node block\"}", true));
+                story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
+
+                List<String> log = p.getLastBuild().getLog(10);
+                String firstNode = "";
+                for (String line: log) {
+                    if (line.contains("Running on")) {
```
jglick: More robust to look for `WorkspaceAction` here.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -486,6 +486,17 @@ public String getCookie() {
             return getDisplayName();
         }
 
+        @Override
+        public String getAffinityKey() {
+            String enclosingLabel = getEnclosingLabel();
```
jglick: That method is designed for UI (like the executor widget) and is not actually what you want here. For example, given

```groovy
node {
  stage('Build') {
    sh 'make'
  }
  stage('Test') {
    sh 'make test'
  }
}
```

the result of `getEnclosingLabel` will vary according to which stage we are currently inside. That is inappropriate for the affinity key, which is supposed to be a stable function of what is _outside_ this `node` block. (After all, the affinity key should be checked just when the `node` block starts.)

So all the machinery with the hash codes and the `match` logic is out of place here, and there is no need to patch `getEnclosingLabel`. You want something much simpler for this purpose: search `executorStepNode.iterateEnclosingBlocks` for some identifier which ought to be stable across runs.

Alternately, just use `ownerTask.name` by default as the affinity key, and if users wish to control the affinity better, let them do so as an explicit parameter to `node`. My comment in https://github.com/jenkinsci/junit-plugin/pull/94#discussion_r155143817 follows similar logic.

orgi: Thanks for the hint about this. Thinking about it a second time I could have just done a much simpler implementation instead of using `getEnclosingLabel()`.
Initially I was only returning  ownerTask,name. However it turned out that when using parallel blocks it happens to happen quite often that still different nodes are picked up.
Therefore I'll go with something else. In fact I already have an implementation for checking for stages and parallel blocks outside nodes blocks used at some different place. I'll probably use that one.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
orgi: @jglick: I got two questions about this proposal:

1. How to find the node name or something similar reliable to check for?
Maybe I'm doing something wrong. However when looking at the `WorkspaceAction` here the `getNode()` method returns an empty string. I'd have assumed that that one would hold the name of the node being used.

```java
    private List<WorkspaceAction> getWorkspaceActions(WorkflowRun workflowRun) {
        FlowGraphWalker walker = new FlowGraphWalker(workflowRun.getExecution());
        List<WorkspaceAction> actions = new ArrayList<WorkspaceAction>();
        for (FlowNode n : walker) {
            WorkspaceAction a = n.getAction(WorkspaceAction.class);
            if (a != null) {
                actions.add(a);
            }
        }
        return actions;
    }
```

in the test I have:
```java
                WorkflowRun run = story.j.assertBuildStatusSuccess(p.scheduleBuild2(0));
                List<WorkspaceAction> workspaceActions = getWorkspaceActions(run);

                String firstNode = null;
                for (WorkspaceAction action: workspaceActions) {
                    if (firstNode == null) {
                        firstNode = action.getNode();
                        firstNode = action.getWorkspace().toComputer().getName();
                    } else {
                        assertEquals(action.getNode(), firstNode);
                    }
                }
```

Both `action.getNode()` as well as `action.getWorkspace().toComputer().getName()` return an empty string here (there's only one action retrieved). Maybe the workspace action retrieved is running on the master?

Not sure whether I am doing something wrong. I'd appreciate any hints that point me to the right direction.

2. What to do with the remaining tests
The other tests currently are searching for the node block where a particular echo step has been executed in. I was thinking about writing some file into the workspace in order to check whether the file is still there after restarting the job. But somehow this sounds little bit too complicated. Would you go for parsing the flow nodes here as well?

pom.xml
orgi: Thanks for the hint. Next time I'll take care. For this time I'll change the dependency to 2.145 assuming it won't need the TODO.

orgi:
@jglick: The latest update should address all the mentioned issues. In fact I referenced the released Jenkins version in `pom.xml` and therefore left out the TODO about the Jenkins dependency.
In addition I removed one test as it felt duplicate to me and the tests for the node reuse take some time to complete.

@dwnusbaum: Do we need to do any changes to the `Jenkinsfile`? Currently the check against Jenkins version `2.121.1` is still removed. Is there a need to check against `2.145`?

Please let me know if there's still something to be done for this fix. I'd be happy seeing this merged.

orgi:
@jglick & @dwnusbaum: Ok to merge - or is there still something missing?

orgi:
@dwnusbaum: Merged in the latest changes. I assumed that this would trigger a new build for this PR. However this doesn't seem to be the case. Now I'm, hoping that this very comment will trigger a new build...

dwnusbaum:
@orgi GitHub is having an outage right now, so webhook delivery is paused (and so no new builds will run until the next repository scan ([scan log](https://ci.jenkins.io/job/Plugins/job/workflow-durable-task-step-plugin/indexing/console))), see https://blog.github.com/2018-10-21-october21-incident-report/ and https://status.github.com/messages. I'm not sure if the hook will get delivered eventually or if you'll have to close/open the PR to trigger a new build once GitHub's issues are fixed.

orgi:
@dwnusbaum: Thanks for the clarification. Seems like my update got triggered automatically :). I understand you've been busy with the 2.24 release. Is there anything still blocking the merge of my changes?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -486,6 +486,68 @@ public String getCookie() {
...
+            }
+            return labelName.toString();
+        }
+
+        @Override
+        public String getAffinityKey() {
+            String enclosingLabels = concatenateAllEnclosingLabels();
+            String ownerTaskName = getOwnerTask().getName();
+            if (enclosingLabels != null) {
+                return ownerTaskName + enclosingLabels;
```
dwnusbaum: To make sure I understand, what would the affinity key be for the node step in the following Pipeline, assuming the WorkflowJob is named `Test`:

    stage('Stage1') {
        parallel 'Branch1': {
            node ('Agent1') {
            }
        }, 'Branch2': {
        }
    }

`TestBranch1#Stage1`? `TestStage1#Branch1`? Something else entirely?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -688,6 +714,271 @@ public boolean apply(@Nullable FlowNode input) {
...
+        });
+    }
+
+    private Map<String, String> mapNodeNameToLogText(WorkflowRun workflowRun) throws java.io.IOException{
+        FlowGraphWalker walker = new FlowGraphWalker(workflowRun.getExecution());
+        Map<String, String> nodeNameToLogText = new HashMap<>();
+        StringBuilder previousLogText = new StringBuilder();
+        for (FlowNode n : walker) {
+            WorkspaceAction a = n.getAction(WorkspaceAction.class);
+            LogActionImpl l = n.getAction(LogActionImpl.class);
```
dwnusbaum: This is the old way of storing log data, so now that the workflow-job plugin used in testing is version 2.26 (which switched to `LogStorageAction`), the log text for each node is always just `""`, which seems unintentional. I think you should replace it with `LogAction.class` to avoid depending on the specific implementation.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.support.steps;
 
 import com.google.common.base.Predicate;
+import com.sun.corba.se.spi.orbutil.threadpool.Work;
```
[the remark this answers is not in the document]
orgi: Must be, thanks, I'll remove that.

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.23</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.121.1</jenkins.version>
+        <jenkins.version>2.145</jenkins.version>
```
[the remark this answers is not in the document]
orgi: That sounds like a great idea - didn't think of it yet. However I'm not sure how then to run those tests against 2.145+? Thought that the test harness will always pick the Jenkins version as given in the pom.xml?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -486,6 +486,68 @@ public String getCookie() {
...
+                    labelName.append("#");
+                    labelName.append(currentLabelName);
+                }
+            }
+            return labelName.toString();
+        }
+
+        @Override
+        public String getAffinityKey() {
+            String enclosingLabels = concatenateAllEnclosingLabels();
```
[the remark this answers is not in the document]
orgi: To my understanding this should only be called once at the start for a `node() {}` block.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
orgi: Thanks, I'll adjust this accordingly.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -486,6 +486,68 @@ public String getCookie() {
...
+                return null;
+            }
+
+            StringBuilder labelName = new StringBuilder();
+            for(FlowNode node: executorStepNode.getEnclosingBlocks()) {
+                String threadName = findThreadName(node);
+                String stageName = findStageName(node);
+                String currentLabelName = threadName != null ? threadName: stageName;
+                if (threadName != null && stageName != null) {
+                    LOGGER.log(FINE, "\"findLabelAndParseLog(): Got both thread: {0} and stage: $stageName. Will take thread name ({0}).", threadName);
```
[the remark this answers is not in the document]
orgi: Sure :)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
orgi: This depends on the implementation of `getEnclosingBlocks()`. Currently it would return the inner blocks first - in that case that should be: `TestBranch1#Stage1`. However from functionality point of view it doesn't really matter. What is important is to have all of the block names & the job name in the key. Maybe for consistency I should either remove the '#' or add it between the job name and first enclosing block? I'll have some thought's on this one.

pom.xml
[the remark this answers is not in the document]
dwnusbaum: Take a look at the possible parameters for `buildPlugin` [here](https://github.com/jenkins-infra/pipeline-library#optional-arguments). If you have `buildPlugin(jenkinsVersions: [null, '2.145'])`, then the maven commands run will be something like the following:

    mvn clean install 
    mvn clean install -Djenkins.version=2.145

The property in the second version will overrride the value in the pom, allowing you test against a different Jenkins version.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: Yeah, I agree that the specific format doesn't really matter as long as everything is in it, was mostly just curious. I initially thought it might be important to put a `#` between each enclosing label and the name to avoid unintentional matches, but I can't actually think of any cases that would cause problems, so it seems fine to me.

pom.xml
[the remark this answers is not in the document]
orgi: Thanks for the insight. Works as described.
I just pushed my latest changes after merging in the official master branch.
However for merging my changes to the Jenkinsfile I'll need your help.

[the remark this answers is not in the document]
orgi: Btw.: Are there any plans to support changes to the Jenkinsfile like - let's say - in case there are changes that there will be two builds: One build as currently with the unchanged Jenkinsfile and the other one incorporating the changes?

orgi:
Wonder whether this very last fail has something to do with my changes. Unfortunately I cannot check the junit xml created as I do not have access to the workspace. However maybe this commit will trigger another build which hopefully runs fine - as the previous builds did.

orgi:
I'll try to close / open the PR to trigger a new build.

orgi:
To trigger new build

orgi:
@dwnusbaum: Build's green now. Hope there's nothing left to be done. Please let me know. Thanks.

dwnusbaum:
@orgi Great! I will try to take a look in the next few days.

orgi:
@dwnusbaum: Some days ago I thought I can make the tests little bit faster & more reliable (removing the `sleep()`s. Hope that would not cause additional effort for the review on your side.

orgi:
@dwnusbaum: Small update: Got rid of the `sleep`s in the tests and reverted the changes in `getEnclosingLabel()` as suggested by @jglick. Must have missed that one until now. Finally I think this one is in a good shape for merging. Would be great if you could have a look wthin the next days.

orgi:
@dwnusbaum / @jglick It now has been almost a month since the last update. Please let me know whether I can to do something in addition to get this PR merged. If it helps I'd merge in current master and update the PR accordingly.
So far I can tell that this one is working perfectly fine on our Jenkins. Finally the nodes are getting reused as much as possible.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -739,6 +747,343 @@ public boolean apply(@Nullable FlowNode input) {
...
+        }
+        return actions;
+    }
+
+    /**
+     * Needs at least Jenkins version 2.145 which supports {@link ExecutorStepExecution.PlaceholderTask#getAffinityKey()}
+     */
+    @Issue("JENKINS-36547")
+    @Test public void reuseNodeFromPreviousRun() {
+        story.addStep(new Statement() {
```
jglick: BTW `.then` is more friendly to Java 8.

```java
story.then(r -> {
    DumbSlave slave = r.createOnlineSlave();
    // etc.
});

Comment (orgi) [reply src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java]: Thanks for mentioning. Indeed this looks much better. I'd propose to change it together with any possible additional changes requested.

**Comment by orgi** [comment]
If there's no more other review comment's I'll proceed and implement the change suggested by @jglick. Hope that you'll be in turn finally able to merge this PR.

**Comment by orgi** [comment]
@dwnusbaum: In addition I propose to depend on Jenkins 2.150.1 in order to get rid of the Jenkinsfile adaptation and conditional testing. Do you agree?

File: src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java [hunk 11]
```diff
@@ -480,6 +480,89 @@ public String getCookie() {
...
+            ThreadNameAction tna = flowNode.getPersistentAction(ThreadNameAction.class);
+
+            if (tna != null) {
+                return tna.getThreadName();
+            }
+            return null;
+        }
+
+        static String findStageName(FlowNode flowNode){
+            LabelAction la = flowNode.getPersistentAction(LabelAction.class);
```
jglick: Note that `LabelAction` is used more broadly than just for `stage`, so the method name might be misleading.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -480,6 +480,89 @@ public String getCookie() {
...
+                return null;
+            }
+
+            StringBuilder labelName = new StringBuilder();
+            for(FlowNode node: executorStepNode.getEnclosingBlocks()) {
+                String threadName = findThreadName(node);
+                String stageName = findStageName(node);
+                String currentLabelName = threadName != null ? threadName: stageName;
+                if (threadName != null && stageName != null) {
+                    LOGGER.log(FINE, "\"concatenateAllEnclosingLabels(): Got both thread: {0} and stage: {1}. Will take thread name ({0}).",
```
jglick: IIRC `parallel` blocks also set `LabelAction`, so it would suffice to ignore `ThreadNameAction`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -480,6 +480,89 @@ public String getCookie() {
...
+        /*
+        * TODO: @Override
+        * once the core dependency was updated to a version >= 2.145 (preferably LTS)
+        * the @Override annotation shall be added
+        */
+        public String getAffinityKey() {
+            String enclosingLabels = concatenateAllEnclosingLabels();
+            String ownerTaskName = getOwnerTask().getName();
+            if (enclosingLabels != null) {
+                return ownerTaskName + enclosingLabels;
```
jglick: Perhaps simpler to pass `new StringBuilder(ownerTaskName)` into `concatenateAllEnclosingLabels`. It can just `return` early if something goes wrong.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -739,6 +747,326 @@ public boolean apply(@Nullable FlowNode input) {
...
+                        List<? extends BlockStartNode> enclosingBlocks = atomNode.getEnclosingBlocks();
+                        for (BlockStartNode startNode : enclosingBlocks) {
+                            WorkspaceAction a = startNode.getAction(WorkspaceAction.class);
+                            if (a != null) {
+                                String nodeName = a.getNode();
+                                if (!workspaceActionToLogText.containsKey(nodeName)) {
+                                    workspaceActionToLogText.put(nodeName, new StringWriter());
+                                }
+                                StringWriter writer = workspaceActionToLogText.get(nodeName);
+                                l.getLogText().writeLogTo(0, writer);
```
jglick: Wait, why are we writing to the build log from inside a test?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -739,6 +747,326 @@ public boolean apply(@Nullable FlowNode input) {
...
+    @Test public void reuseNodeInSameRun() {
+        story.then(r -> {
+            for (int i = 0; i < 5; ++i) {
+                DumbSlave slave = r.createOnlineSlave();
+                slave.setLabelString("foo");
+            }
+
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "demo");
+            p.setDefinition(new CpsFlowDefinition("for (int i = 0; i < 20; ++i) {node('foo') {echo \"ran node block ${i}\"}}", true));
+            WorkflowRun run = r.assertBuildStatusSuccess(p.scheduleBuild2(0));
```
jglick:
```suggestion
            WorkflowRun run = r.buildAndAssertSuccess(p);
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
orgi: @jglick: That means we can rely on that? That'd simplify the code. 👍

orgi: I can remember that for some reason I chose to use `getThreadName()` but I cannot recall. However this was in some other code where the exact name did matter. Think in this context it'd not matter at all.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
orgi: @jglick: Actually what happens here is reading the build log and storing it inside `writer`. Or was it that what you were asking for?
This helper just provides the build log to the tests so it can be checked for correctness. But maybe there's a better way of doing that? So far I could not come up with a better idea.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
orgi: Thanks for the hint. Done.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
orgi: Did the proposed change. Looks much simpler now.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
orgi: Done.

orgi:
@dwnusbaum, @jglick : Thanks for the approval & comments. As I usually want to finalized what I started and like to take the opportunity to learn something new :) Therefore I took care about the nits mentioned.
A merge will be highly appreciated as we then finally can get rid of our custom build.

Can you already give a best guess when the other release will be ready and you'll be able to do the merge?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
jglick: `JenkinsRule.assertLogContains`?

orgi: Thanks. Actually I forgot to mention that I need to check for the console log which belongs to a specific node block to see whether it has been executed on the expected node. First think I tried was to somehow retrieve the node name and check that.

After a while I thought it'd be better to have a small helper which will split the console log and map the node names to their corresponding log text.
Now that I think about it I probably could have also just retrieved the node name using `getContext()` and `echo` it to the console where required.

However since the solution in place will be little bit more flexible I'd prefer that one.
I added some doxygen comment to hopefully clarify about that method's purpose.

Ok to keep it?

orgi:
That's great news, thanks 😄

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -480,6 +480,71 @@ public String getCookie() {
...
+         *
+         * In case there's no context available or we get a timeout we'll just return <code>baseLabel</code>
+         *
+         * */
+        private String concatenateAllEnclosingLabels(StringBuilder labelName) {
+            if (!context.isReady()) {
+                return labelName.toString();
+            }
+            FlowNode executorStepNode = null;
+            try (Timeout t = Timeout.limit(100, TimeUnit.MILLISECONDS)) {
```
darxriggs: When or how often is it expected to run into this timeout?
I am just wondering if this will result in many thrown exceptions on a system under load put it under even more stress.

orgi: The timeout was taken from the code used to find the enclosing label for displaying in the UI.
I do not expect it to happen frequently. We have a big system (node allocation itself sometimes takes more than a minute - even for Freestyle jobs and pipeline jobs BEFORE applying this patch) and never got any timeout here.
Hope that helps.
Since this was done analogous to what is done for the UI display maybe one of the maintainers can comment on this, too?

dwnusbaum: AFAIU the timeout was added to the UI code in [this commit](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/2/commits/771d3aa57e65330b703e82abe2f9835471676f71) because `context.get(FlowNode.class)` could block before [this change](https://github.com/jenkinsci/workflow-cps-plugin/pull/145) in workflow-cps. Based on a quick read of the discussion in #2 I think the call is expected to be instantaneous, see [this comment](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/2#discussion_r122758426). Maybe the timeout here is unnecessary.

jglick: For UI code, certainly a timeout is in order—the exact result is not that important, we just need to complete the HTTP response, _now_.

For this code I am less sure, but I think it would be called in the context of `Queue.maintain`, and that also is expected to run in subsecond times. Whether a long delay is even theoretically possible here, I am not sure offhand; would have to take time to study it.

PR #80, opened 2018-10-18
[JENKINS-54133] Document remote console note behavior in test

jglick:
Studying [JENKINS-54133](https://issues.jenkins-ci.org/browse/JENKINS-54133). ~Safe to merge, but does not do anything at the moment.~

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
@@ -10,13 +10,18 @@ import hudson.EnvVars;
 import hudson.Functions;
 import hudson.Launcher;
 import hudson.LauncherDecorator;
+import hudson.MarkupText;
 import hudson.console.AnnotatedLargeText;
+import hudson.console.ConsoleAnnotator;
+import hudson.console.ConsoleLogFilter;
+import hudson.console.ConsoleNote;
 import hudson.console.LineTransformationOutputStream;
 import hudson.model.BallColor;
 import hudson.model.BuildListener;
 import hudson.model.FreeStyleProject;
 import hudson.model.Node;
 import hudson.model.Result;
+import hudson.model.Run;
 import hudson.remoting.Channel;
 import hudson.model.Slave;
 import hudson.model.TaskListener;
@@ -30,10 +35,13 @@ import java.io.IOException;
 import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.Serializable;
+import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.logging.Level;
 import java.util.logging.LogRecord;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -459,6 +475,102 @@ private Object writeReplace() {
...
+            }
+        }
+        private static final class Filter extends ConsoleLogFilter implements Serializable {
+            private final @CheckForNull byte[] note;
+            Filter(boolean smart) throws IOException {
+                JenkinsJVM.checkJenkinsJVM();
+                if (smart) {
+                    ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                    new HelloNote().encodeTo(baos);
+                    note = baos.toByteArray();
```
dwnusbaum: The `smart` approach feels terribly hacky. If we are going to use it in actual code as in jenkinsci/ant-plugin#32, is there any documentation we can update (maybe `ConsoleLogFilter` or `ConsoleNote` itself) to clearly explain the problem with using `ConsoleNote` remotely, describe the pre-generation hack, and describe the preferred way to accomplish the same result for new code (Which I think is `ConsoleAnnotatorFactory` if I understand [your comment on JENKINS-54133](https://issues.jenkins-ci.org/browse/JENKINS-54133?focusedCommentId=351673&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-351673) correctly.)

jglick: Documented in https://github.com/jenkinsci/jenkins/pull/3703/commits/8d73038a27420fdee0f5b4f97b47229c243345e8.

jglick: (prepared in a core PR)

dwnusbaum: https://github.com/jenkinsci/jenkins/pull/3703 for a link to the PR.

jglick:
> The `smart` approach feels terribly hacky.

Perhaps. But consider the `ant` example, and suppose that we are generating a gigantic log from an Ant build (such as would happen when you type `ant` in the NetBeans monorepo, for example) and logs are being streamed from an agent to some external sink. Using solely a `ConsoleAnnotatorFactory` is not really practical because these apply globally to all parts of all builds, and this is matching a rather generic pattern like the three lines consisting of target names in

```
…
jar-ml:

netbeans:
  [genlist] Generating Auto Update information for org.openide.awt
   [repeat] 
   [repeat] all-api.progress:
     [echo] Building platform/api.progress...

init-tasks:
…
```

that is only identifiable as “Ant target names” due to the context that this is coming from a process running Ant and not something completely unrelated. So if we go the `ConsoleNote` route, we need to emit these notes from the agent side. But we cannot trust the agent JVM to sign notes in general, because they are often (though not in this case) written to render instance-supplied HTML or other XSS-oriented data, and we assume that any agent JVM might have been compromised.

So, unless a very different approach is taken, we need to have the master JVM create and sign the notes in advance, then serve them to the agent callable.

dwnusbaum:
I agree that pre-generating the notes makes sense given the constraints, it just seems like the API was not really designed to support this use case cleanly. Given that it's the best option in some cases, I think updating the documentation like you did will be very helpful in case it comes up for anyone in the future.

jglick:
> [It] seems like the API was not really designed to support this use case cleanly

Indeed. It is a bit hard to say what the API _was_ designed for. See https://github.com/jenkinsci/jenkins/commit/2d7e0d96b03ff4aeba4a27bc1dec1355c59fe91c. There is a more detailed discussion in SECURITY-382.

PR #81, opened 2018-10-19
Reverting custom thread pool from #53 in watch mode

jglick:
Observed to cause anomalous behavior during a `RestartableJenkinsRule` test in the `ant` plugin after updating dependencies in https://github.com/jenkinsci/ant-plugin/pull/32. I think this is because the thread pool is not getting shut down, so two copies of the build get loaded after restart and mayhem ensues. Most likely this would not happen in production systems, which should be throwing out the plugin class loaders across restarts. Anyway as of #63 we make far fewer `check` calls, so the rationale for avoiding the shared `Timer` in #53 is obsolete.


jglick:
> We have already hit issues in the past with overloading the Timer thread pool from Pipeline

Because it was being used for purposes which now it is not.

> 4. It's not clear that the original issue is confirmed to come from this cause in the first place.

It is pretty clear to me. The test failed until I introduced this fix, then it passed.

jglick:
> it is unfortunately _not_ so configured (perhaps a mistake on our part)

Possibly. I seem to recall issues enabling that. Could be revisited.

> assuming 1 MB stack size and 10 threads, 10 MB even if only 1-2 tasks at a time are usually running

Exactly why it is undesirable to introduce _additional_ thread pools for plugin use when they are not strictly necessary: reusing an existing pool minimizes overhead.

jglick:
> this _could_ be trivially (and more correctly) handled via a `Terminator` or explicit shutdown hook

I am willing to do that if it helps get https://github.com/jenkinsci/ant-plugin/pull/32 out the door, though I still think this PR is preferable, at least once #84 is reverted.

jglick:
On hold because of #84.

svanoort:
> It is pretty clear to me. The test failed until I introduced this fix, then it passed.

Ah, okay, well that would have been good to know in the PR description.  

> Exactly why it is undesirable to introduce additional thread pools for plugin use when they are not strictly necessary: reusing an existing pool minimizes overhead.

I don't disagree with *generally* consolidating pools, just want to make we don't have thread-pool exhaustion risks.  Revising that pool implementation would sort that out just fine.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -205,12 +202,6 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
         private static final long MAX_RECURRENCE_PERIOD = 15000; // 15s
         private static final float RECURRENCE_PERIOD_BACKOFF = 1.2f;
 
-        private static final ScheduledThreadPoolExecutor THREAD_POOL = new ScheduledThreadPoolExecutor(25, new NamingThreadFactory(new DaemonThreadFactory(), DurableTaskStep.class.getName()));
```
jglick: Needs to be made sensitive to `USE_WATCHING`.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -210,7 +211,10 @@ public FormValidation doCheckLabel(@QueryParameter String label) {
     public static long REMOTE_TIMEOUT = Integer.parseInt(System.getProperty(DurableTaskStep.class.getName() + ".REMOTE_TIMEOUT", "20"));
 
     private static ScheduledThreadPoolExecutor threadPool;
-    private static synchronized ScheduledThreadPoolExecutor threadPool() {
+    private static synchronized ScheduledExecutorService threadPool() {
+        if (USE_WATCHING) {
```
jglick, line 215: Note that this feature flag remains off by default.

car-roll (approved):

PR #82, opened 2018-10-22
Update to non-beta versions of dependencies

dwnusbaum:
These dependencies were added by #21.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
         <jenkins.version>2.121.1</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.13</workflow-step-api-plugin.version>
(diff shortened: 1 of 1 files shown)


dwnusbaum:
Attempting to retrigger the build.

pom.xml
```diff
@@ -99,7 +99,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.26-beta-1</version>
+            <version>2.26</version>
```
dwnusbaum: This version is blacklisted on the update center right now, but the dependency is test-scope only and the Maven release is totally fine, so it shouldn't cause any issues.

PR #83, opened 2018-10-23
Fix commit hashes in test comments

dwnusbaum:
See https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75#discussion_r227280773.

@kutzi noticed that the previous commit hashes here were not correct. In fact, they were from a different repo: https://github.com/jenkinsci/durable-task-plugin/commit/a224295d12. A commit in this repo that demonstrates the previous behavior is 3bd3498ba3, and I just confirmed locally that if you check out that commit and add the new tests, they fail as described. Maybe pointless to reference commits in the first place, since it probably isn't useful to anyone, but it won't hurt either.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
@@ -522,7 +522,7 @@ public class ShellStepTest {
                 "}", true));
         WorkflowRun b = p.scheduleBuild2(0).waitForStart();
         j.waitForCompletion(b);
-        // Fails with Result.FAILURE as of commit a224295d12.
+        // Fails with Result.FAILURE as of commit 3bd3498ba3.
         j.assertBuildStatus(Result.ABORTED, b);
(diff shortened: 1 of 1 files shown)


kutzi:
Okay, thanks for tackling this.
But maybe I don't understand the comments then at all.
Why does it e.g. say `Fails with Result.FAILURE` when then it's asserted that the status is ABORTED?

dwnusbaum:
@kutzi The point is to explain what the behavior was before the fix.

kutzi:
But then the wording sounds strange to me - though I'm not a native speaker.
Did you want to express something like `would fail with Result.FAILURE until commit ...`?

kutzi:
Anyway, IMHO the comments do more to confuse the reader than anything else. They are really only of historic interest and bring nothing in kind of understanding the test better.

dwnusbaum:
@kutzi I tried to clarify the comments. To me, the whole point of these tests is to show how the behavior changed as a result of the PR, so a comment for what the previous behavior was seems useful to anyone investigating related bugs in the future.

kutzi:
Okay, that makes it clearer for me

PR #84, opened 2018-10-24
[JENKINS-52165] Disable watch mode by default for now

jglick:
Disabling #63 to buy some time until I can fix some acknowledged issues (mainly [JENKINS-54133](https://issues.jenkins-ci.org/browse/JENKINS-54133) and waiting for the maintainer on [JENKINS-54081](https://issues.jenkins-ci.org/browse/JENKINS-54081)) and properly investigate some reported issues which are likely linked ([JENKINS-54073](https://issues.jenkins-ci.org/browse/JENKINS-54073) and [JENKINS-53888](https://issues.jenkins-ci.org/browse/JENKINS-53888)). Complements https://github.com/jenkins-infra/update-center2/pull/247. https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/74 is still needed promptly since that seems to be independent.

Note that this retains watch mode in tests, to continue to enforce coverage in this plugin and in any other pkugins which currently have a test dep on 2.22+ which might bump up their dep to 2.25.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -31,6 +31,7 @@ import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Functions;
 import hudson.Launcher;
+import hudson.Main;
 import hudson.Util;
 import hudson.model.Node;
 import hudson.model.TaskListener;
@@ -170,7 +171,7 @@ public abstract class DurableTaskStep extends Step {
     /** If set to false, disables {@link Execution#watching} mode. */
     @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "public & mutable only for tests")
     @Restricted(NoExternalUse.class)
-    public static boolean USE_WATCHING = !"false".equals(System.getProperty(DurableTaskStep.class.getName() + ".USE_WATCHING"));
+    public static boolean USE_WATCHING = Boolean.parseBoolean(System.getProperty(DurableTaskStep.class.getName() + ".USE_WATCHING", Main.isUnitTest ? "true" : /* JENKINS-52165 turn back on by default */ "false"));
 
     /**
      * Represents one task that is believed to still be running.
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -467,14 +467,7 @@ public abstract class DurableTaskStep extends Step {
                             if (controller.writeLog(workspace, listener.getLogger())) {
                                 LOGGER.log(Level.FINE, "last-minute output in {0} on {1}", new Object[] {remote, node});
                             }
-                            if (returnStatus || exitCode == 0) {
-                                getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(controller.getOutput(workspace, launcher()), StandardCharsets.UTF_8) : null);
-                            } else {
-                                if (returnStdout) {
-                                    listener.getLogger().write(controller.getOutput(workspace, launcher())); // diagnostic
-                                }
-                                getContext().onFailure(new AbortException("script returned exit code " + exitCode));
-                            }
+                            handleExit(exitCode, () -> controller.getOutput(workspace, launcher()));
                             recurrencePeriod = 0;
                             controller.cleanup(workspace);
                         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -170,7 +171,7 @@ public FormValidation doCheckReturnStatus(@QueryParameter boolean returnStdout,
     /** If set to false, disables {@link Execution#watching} mode. */
     @SuppressFBWarnings(value = "MS_SHOULD_BE_FINAL", justification = "public & mutable only for tests")
     @Restricted(NoExternalUse.class)
-    public static boolean USE_WATCHING = !"false".equals(System.getProperty(DurableTaskStep.class.getName() + ".USE_WATCHING"));
+    public static boolean USE_WATCHING = Boolean.parseBoolean(System.getProperty(DurableTaskStep.class.getName() + ".USE_WATCHING", Main.isUnitTest ? "true" : /* JENKINS-52165 turn back on by default */ "false"));
```
dwnusbaum: It seems undesirable to not actually test using the mode that we are shipping to users, but I also see why it would be good to test against watch mode since any issues in downstream plugins are likely to be specific to that mode, so I am somewhat split on this.

jglick: > It seems undesirable to not actually test using the mode that we are shipping to users

Indeed. On the other hand the non-watch mode has been tested for years, and I certainly hope this situation will be temporary.

dwnusbaum:
I was curious to see what tests failed if I did not use watch mode for tests, and among the expected failures related to remote logging, the [tests](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/593ea9f885f4f8a10129b78295647baf113fe4f8/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java#L513-L545) for my recent JENKINS-28822 fix started failing, because I only fixed it for watch mode. It would be ideal to modify that fix to work for non-watch mode as well, or we'd need to reopen the issue.

jglick:
> my recent JENKINS-28822 fix

#75 FTR. Indeed, that fix needs to be extended to work in both modes. I will check it.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -541,7 +541,7 @@ private Object writeReplace() {
         j.waitForCompletion(b);
         // Would have succeeded before https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75.
         j.assertBuildStatus(Result.ABORTED, b);
-        j.assertLogContains("Timeout has been exceeded", b);
+        j.waitForMessage("Timeout has been exceeded", b); // TODO assertLogContains fails unless a sleep is introduced; possible race condition in waitForCompletion
```
dwnusbaum: I saw this as well, although it passed sometimes, so a race condition seems likely.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
svanoort: I'm okay with this part -- not perfect but it'll do.  Just trying to check that the exit condition stuff is 100% the same internally.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -507,12 +501,20 @@ private void check() {
...
+
+        @FunctionalInterface
+        private interface OutputSupplier {
+            byte[] produce() throws IOException, InterruptedException;
+        }
+        private void handleExit(int exitCode, OutputSupplier output) throws IOException, InterruptedException {
             Throwable originalCause = causeOfStoppage;
             if ((returnStatus && originalCause == null) || exitCode == 0) {
-                getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(output, StandardCharsets.UTF_8) : null);
+                getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(output.produce(), StandardCharsets.UTF_8) : null);
```
svanoort: :ant: Nested ternary operators are super hard to read for most developers without line breaks between them

jglick: Perhaps. Anyway that predates this PR.

svanoort:
Double approval *could* be because I approve extra hard... or because conference wifi sucks.  You decide!

PR #86, opened 2018-10-26
Integration tests against buffered output

jglick:
Was originally going into #85 until that got merged. See https://github.com/jenkinsci/workflow-api-plugin/pull/81.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -282,6 +282,7 @@ public class ExecutorStepTest {
     @Issue("JENKINS-52165")
     @Test public void shellOutputAcrossRestart() throws Exception {
         Assume.assumeFalse("TODO not sure how to write a corresponding batch script", Functions.isWindows());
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -282,6 +282,7 @@ private void startJnlpProc() throws Exception {
     @Issue("JENKINS-52165")
     @Test public void shellOutputAcrossRestart() throws Exception {
         Assume.assumeFalse("TODO not sure how to write a corresponding batch script", Functions.isWindows());
+        Assume.assumeThat("TODO no longer asserts anything, just informational. There is no way for FileMonitoringTask.Watcher to know when content has been written through to the sink other than by periodically flushing output and declining to write lastLocation until after this completes. This applies both to buffered on-master logs, and to typical cloud sinks.", System.getenv("JENKINS_URL"), nullValue());
```
dwnusbaum: It seems like it could still be useful to assert that you haven't lost a huge amount of output, say 5% or higher.

jglick:
I am still thinking about ways that `shellOutputAcrossRestart` could be made to pass again, at least with some restrictions. For example if you were doing a graceful shutdown of the master, could a `@Terminator` be made to flush content? Or the previous idea of only writing `lastLocation` after some occasional flushes.

jglick:
OTOH this is safe to merge simply as a way to get clean PCT runs.

jglick:
> ways that `shellOutputAcrossRestart` could be made to pass again

Filed as [JENKINS-56851](https://issues.jenkins-ci.org/browse/JENKINS-56851).

PR #89, opened 2018-11-01
No need to override default impls from AccessControlled

jglick:
Amends @roguishmountain’s #68 using https://github.com/jenkinsci/jenkins/pull/2999.


dwnusbaum:
The CI build appears to be gone, so I am trying to trigger a new build.

PR #92, opened 2018-12-14
[JENKINS-52943] Add descriptors for shell steps

halkeye:
See [JENKINS-52943](https://issues.jenkins-ci.org/browse/JENKINS-52943).

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
@@ -53,7 +53,6 @@ import jenkins.util.JenkinsJVM;
 import org.apache.commons.io.FileUtils;
 import static org.hamcrest.Matchers.*;
 
-import org.hamcrest.Matchers;
 import org.jenkinsci.plugins.workflow.actions.ArgumentsAction;
 import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
 import org.jenkinsci.plugins.workflow.cps.CpsStepContext;
@@ -135,9 +134,7 @@ public class ShellStepTest {
         assertTrue(found);
     }
 
-    /**
-     * Failure in the shell script should mark the step as red
-     */
+    @Issue("JENKINS-52943")
     @Test
     public void stepDescription() throws Exception {
         // job setup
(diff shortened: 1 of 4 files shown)


halkeye:
WIP because I'm trying to learn how to write a test for this.

I did test it manually though.

image

Generated by the script:
```
node('master'){
 sh 'echo "hello"'
 sh returnStdout: true, script: 'echo "hello"'
 sh returnStatus: true, script: 'echo "hello"'
}
```

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -128,6 +135,34 @@ public void failureShouldMarkNodeRed() throws Exception {
...
+        WorkflowRun b = j.buildAndAssertSuccess(foo);
+
+        ArrayList<String> args = new ArrayList<>();
+        List<FlowNode> shellStepNodes = new DepthFirstScanner().filteredNodes(b.getExecution(), new NodeStepTypePredicate(Functions.isWindows() ? "bat" : "sh"));
+        assertThat(shellStepNodes, Matchers.hasSize(2));
+        for(FlowNode n : shellStepNodes) {
+            args.add(ArgumentsAction.getStepArgumentsAsString(n));
+        }
+
+        assertArrayEquals(new String[] { "first", "second" }, args.toArray());
```
jglick: BTW there is a more idiomatic way of doing this with `Matchers`, something like

```java
assertThat(args, contains("first", "second"));
```

but getting the `static import`s and generics exactly right can be tricky if you are not really comfortable with Java 5+.

jglick:
Your problem is that this plugin was still testing against a very old version of `workflow-cps`. The API here was [implemented in 2.31](https://plugins.jenkins.io/workflow-cps); older versions will not save step arguments. Minimally:

```diff
diff --git a/pom.xml b/pom.xml
index 1987911..4b53c92 100644
--- a/pom.xml
+++ b/pom.xml
@@ -93,7 +93,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.22</version>
+            <version>2.31</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -137,7 +137,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-scm-step</artifactId>
-            <version>2.3</version>
+            <version>2.4</version>
             <scope>test</scope>
         </dependency>
         <dependency>
```

(You could go higher, all the way up to current 2.61, though that might pull in some other newer dependencies too which would necessitate larger POM changes.)

halkeye:
thanks @dwnusbaum

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
halkeye: so @dwnusbaum provided a PR that fixed that too.

PR #93, opened 2019-01-04
[JENKINS-55410] Added label attribute for documentation and clarity in jenkins views

soenkekueper:
See [JENKINS-55410](https://issues.jenkins-ci.org/browse/JENKINS-55410).

I've added an new Attribute "label" to the durable steps, so that this is displayed within the pipeline steps view and blue ocean views instead of the technical Name "Windows Batch Script" or "Linux Shell script". We have got a lot of this and it is very hard to find the right one. 

With this labels this step can now display the real domain specific information for example "build manual" "copy driver files" etc. This makes the pipeline views very more comfortable to use.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -61,6 +61,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.util.Timer;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.durabletask.Controller;
 import org.jenkinsci.plugins.durabletask.DurableTask;
 import org.jenkinsci.plugins.durabletask.Handler;
@@ -93,6 +94,8 @@ import org.kohsuke.stapler.QueryParameter;
 public abstract class DurableTaskStep extends Step {
 
     private static final Logger LOGGER = Logger.getLogger(DurableTaskStep.class.getName());
+
+    private static final int MAX_LABEL_LENGTH = 100;
 
     private boolean returnStdout;
     private String encoding;
@@ -135,7 +138,7 @@ public abstract class DurableTaskStep extends Step {
 
     @Override public StepExecution start(StepContext context) throws Exception {
         if (this.label != null && !this.label.isEmpty()) {
-            context.get(FlowNode.class).addAction(new LabelAction(this.label));
+            context.get(FlowNode.class).addAction(new LabelAction(StringUtils.left(label, MAX_LABEL_LENGTH)));
         }
         return new Execution(context, this);
     }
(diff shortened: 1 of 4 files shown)


dwnusbaum:
Thanks for the PR!

After #92 `sh`/`bat` steps in Blue Ocean should at least show the contents of the script in all cases, see https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/92#issuecomment-447194008 for a screenshot, but it still seems like this would still be a useful feature. 

We try to track new features through Jira tickets, so would you be able to open a ticket [here](https://issues.jenkins-ci.org/secure/Dashboard.jspa) with component `workflow-durable-task-step`, basically copying the description of your PR to describe what the feature is and why you want it?

abayer:
Could you attach screenshots of what this makes Blue Ocean look like? Thanks.

soenkekueper:
Hey,
this pipeline
`node{
    bat label: 'Generate and compile manual', script: 'buildManual.cmd'
}`

procueds this blue-ocean view
labeledsteppipeline

and this step view

labeledstepstepview

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -121,8 +124,19 @@ public boolean isReturnStatus() {
...
+        this.label = label;
+    }
+    
+    public String getLabel() {
+        return label;
+    }
 
     @Override public StepExecution start(StepContext context) throws Exception {
+        if (this.label != null && !this.label.isEmpty()) {
+            context.get(FlowNode.class).addAction(new LabelAction(this.label));
```
abayer: I'd suggest deciding on a maximum length and trimming the label if it's longer than that - otherwise, the UI could get really wonky.

dwnusbaum:
> Which seems to me like any kind of flakeyness. The newly added testcase labelShortened has passed.
> Any ideas whats going wrong? I've no one...

Yeah it seems like something flaky on the CI instance to me. I'll close this PR and reopen it to restart the build.

dwnusbaum:
Thanks for the improvement @soenkekueper!

jglick:
Would it not be better for the `label` to be used as the return value of `argumentsToString` and skip the `LabelAction`?

jglick:
@bitwiseman

> Overriding it in this base class would require changes to the derived classes as well.

Sure, this would be a two-minute refactoring. The question is about the user expectation. I would have expected a label to override the actual arguments, so that if I write, say,

```groovy
sh label: 'Collect frobnitzes', script: '''
mvn -B -Dsome.long.option -Dsome.even.longer.option -f whatever/subdir org.whatever:frobnitz-maven-plugin:1.0.3-beta-99:collect
'''
```

that all compliant UIs will display just **Collect frobnitzes** and not the crazy command line. (Currently, if you omit `set +x` / `@echo off`, the command line will actually get echoed by the shell if you care to look at the step’s log text, so there is no loss of diagnostic ability if something goes wrong.)

jglick:
As noted in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/80, what I actually advocated from the start was something more like

```groovy
stage('Collect frobnitzes') { // or even label('…')
  sh '''
  mvn -B -Dsome.long.option -Dsome.even.longer.option -f whatever/subdir org.whatever:frobnitz-maven-plugin:1.0.3-beta-99:collect
  '''
}
```

which would allow a UI to offer whatever level of detail it could comfortably display—permitting drill-down to individual steps where feasible.

PR #95, opened 2019-01-18
Update dependencies and remove test using LocalData to fix PCT failures

dwnusbaum:
Analogous to jenkinsci/workflow-basic-steps-plugin#79 and jenkinsci/pipeline-input-step-plugin#31.

Most notable change is that I deleted a test using `@LocalData` to verify the serial compatibility of `node`/`sh` steps simply because I could not figure out how to recreate the data in a way that kept the test working with only minor changes (I had tweaked workspace names and similar to line up correctly, but the build just hung when it restarted). 

The test was originally added in https://github.com/jenkinsci/workflow-durable-task-step-plugin/commit/3744da4e1a81e77f74194385191d9e33129ec4ff in early 2017, so the specific change it was a response to is not particularly relevant anymore, but some kind of regression test for the serial form would probably still be desirable. Thoughts? If the consensus is that this type of test is valuable then I'd need to spend more time trying to understand how to update the test to keep it working.


dwnusbaum:
Merging now, we can reintroduce a similar test later if desired.

PR #96, opened 2019-01-23
[JENKINS-52165] Do not enable watch mode automatically in tests

jglick:
Besides the original discussion in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/84#discussion_r227798017, I found that watch mode was being activated in JFR “serverless” builds, including Jenkins X in `--prow` mode, due to [this questionable behavior](https://github.com/jenkinsci/jenkinsfile-runner/blob/6ca2605ea48a4c186ed48d16dcbb5db8551b5b54/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java#L654-L655). Probably less surprising to enable watch mode only when tests explicitly ask for it.

I considered making some of these test suites `Parameterized` but they were already pretty long-running and it looked complicated: some tests only make sense in watch mode or vice-versa, so was not sure whether this would be desirable.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
@@ -25,7 +25,6 @@ import hudson.model.Run;
 import hudson.remoting.Channel;
 import hudson.model.Slave;
 import hudson.model.TaskListener;
-import hudson.remoting.Command;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.EnvironmentVariablesNodeProperty;
 import hudson.tasks.BatchFile;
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -463,114 +464,10 @@ private Object writeReplace() {
         }
     }
 
-    @Ignore("TODO too flaky to run in CI")
-    @Issue("JENKINS-38381")
-    @Test public void remoteVoluminousLogger() throws Exception {
```
jglick: Turned out to be too hard to make this be meaningful in a functional test; giving up on it for now.

jglick:
> due to [this questionable behavior](https://github.com/jenkinsci/jenkinsfile-runner/blob/6ca2605ea48a4c186ed48d16dcbb5db8551b5b54/setup/src/main/java/io/jenkins/jenkinsfile/runner/JenkinsEmbedder.java#L654-L655)

[JENKINS-55764](https://issues.jenkins-ci.org/browse/JENKINS-55764)

PR #100, opened 2019-03-28
Update description to be exposed in the Plugin Manager UI view

v1v:
It shows `The Jenkins Plugins Parent POM Project`, see below:

image

Let's use the short description from https://plugins.jenkins.io/workflow-durable-task-step

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,6 +36,9 @@
     <version>${revision}${changelist}</version>
     <packaging>hpi</packaging>
     <name>Pipeline: Nodes and Processes</name>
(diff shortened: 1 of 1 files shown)


v1v:
Indeed. I've found a bunch of plugins which don't have that particular description. So I took the lead to raise this PR rather than asking more widely in the mailing list. It sounds reasonable the reason for using the aggregator as the main interaction, but it's also awkward to skip the description by default. In both cases, I do not have any strong opinion. Let's see what others think about it.

Thanks @dwnusbaum

PR #70, opened 2018-05-22
Test cases for JENKINS-41854

basil:
# Problem

[JENKINS-41854](https://issues.jenkins-ci.org/browse/JENKINS-41854) affects my production Jenkins servers at least once a month. My only recourse is to restart Jenkins, which rehydrates the `FilePath` as described in the bug. Between the time that I hit the bug and the time I restart Jenkins, in-use workspaces are handed out to new jobs, causing both jobs to fail. I provided a detailed write-up of my experiences in [JENKINS-50504](https://issues.jenkins-ci.org/browse/JENKINS-50504), which I have now marked as a duplicate of JENKINS-41854 (I didn't know about JENKINS-41854 at the time).

# Solution

Work towards fixing the bug by writing a reproducible test case. This test case focuses on highlighting the two failure modes described in the bug:

- If an agent is disconnected and reconnected during the first of two `sh` steps in a node, then the second `sh` step will fail.
- If an agent is disconnected and reconnected, a fresh `WorkspaceList` lock needs to be acquired, after a new `Computer` becomes available and hence a new `WorkspaceList`.

# Implementation

I found that the existing `buildShellScriptAcrossDisconnect` test could be extended to cover this bug, so I added some additional logic to it to demonstrate these two failure modes. I left the new logic commented out in places where it would currently cause the test to fail (until JENKINS-41854 is fixed). An alternative approach would have been to copy and paste the test into a new test, add the new logic there (without commenting it out), and skip the new test using `@Ignore`. Please let me know if you'd prefer that approach instead.

# Testing

When the "Back again" line is uncommented, the test fails as described in the bug (and with the same exception I see on my production Jenkins server with real jobs):

```
   4.634 [demo #1] Started
   5.891 [demo #1] [Pipeline] node
   8.484 [demo #1] Running on dumbo in /var/folders/sd/c1y1b1c542z88kn6k7drmq2h0000gq/T/junit5210374757619203863/workspace/demo
   8.485 [demo #1] [Pipeline] {
   9.915 [demo #1] [Pipeline] sh
  10.608 [demo #1] [demo] Running shell script
  10.608 [demo #1] Cannot contact dumbo: hudson.remoting.RequestAbortedException: java.nio.channels.ClosedChannelException
  12.344 [demo #1] + touch /var/folders/sd/c1y1b1c542z88kn6k7drmq2h0000gq/T/junit6462864648634361788/junit5482144690654488963/f2
[...]

PR #101, opened 2019-03-29
[JENKINS-41854] Recontextualizing FilePath on demand

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/42 & https://github.com/jenkinsci/workflow-cps-plugin/pull/291. Enables test created in #70.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
@@ -51,6 +51,9 @@ public class FilePathPickle extends Pickle {
             protected FilePath tryResolve() {
                 return FilePathUtils.find(slave, path);
             }
+            @Override protected FlowExecutionOwner getOwner() {
+                return owner;
+            }
             @Override public String toString() {
                 return "Looking for path named ‘" + path + "’ on computer named ‘" + slave + "’";
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -69,12 +69,10 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.FlowScanningUtils;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
-import org.jenkinsci.plugins.workflow.steps.DynamicContext;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.durable_task.Messages;
 import org.jenkinsci.plugins.workflow.support.actions.WorkspaceActionImpl;
 import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
-import org.jenkinsci.plugins.workflow.support.pickles.PickleDynamicContext;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -785,7 +783,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         listener.getLogger().println("Running on " + ModelHyperlinkNote.encodeTo(node) + " in " + workspace);
                         context.newBodyInvoker()
                                 .withContexts(exec, computer, env,
-                                    DynamicContext.merge(context.get(DynamicContext.class), new PickleDynamicContext(workspace)))
+                                    FilePathDynamicContext.createContextualObject(workspace))
                                 .withCallback(new Callback(cookie, lease))
                                 .start();
                         LOGGER.log(FINE, "started {0}", cookie);
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -69,10 +69,12 @@ import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.graphanalysis.FlowScanningUtils;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
 import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
+import org.jenkinsci.plugins.workflow.steps.DynamicContext;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.durable_task.Messages;
 import org.jenkinsci.plugins.workflow.support.actions.WorkspaceActionImpl;
 import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
+import org.jenkinsci.plugins.workflow.support.pickles.PickleDynamicContext;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -782,7 +784,8 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         }
                         listener.getLogger().println("Running on " + ModelHyperlinkNote.encodeTo(node) + " in " + workspace);
                         context.newBodyInvoker()
-                                .withContexts(exec, computer, env, workspace)
+                                .withContexts(exec, computer, env,
+                                    DynamicContext.merge(context.get(DynamicContext.class), new PickleDynamicContext(workspace)))
                                 .withCallback(new Callback(cookie, lease))
                                 .start();
                         LOGGER.log(FINE, "started {0}", cookie);
--- a/pom.xml
+++ b/pom.xml
@@ -66,8 +66,8 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
-        <workflow-step-api-plugin.version>2.18</workflow-step-api-plugin.version>
-        <workflow-support-plugin.version>3.1</workflow-support-plugin.version>
+        <workflow-step-api-plugin.version>2.20-rc519.c38481e3078b</workflow-step-api-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-step-api-plugin/pull/42 -->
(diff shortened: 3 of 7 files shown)


jglick:
Our Windows agents are very flaky…

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -782,7 +782,8 @@ public void run() {
                         }
                         listener.getLogger().println("Running on " + ModelHyperlinkNote.encodeTo(node) + " in " + workspace);
                         context.newBodyInvoker()
-                                .withContexts(exec, computer, env, workspace)
+                                .withContexts(exec, computer, env,
+                                    FilePathDynamicContext.createContextualObject(workspace))
```
dwnusbaum: Since `DynamicContext.get` takes precedence over `StepContext.doGet`, what happens after this patch for something like:

```
node {
  dir('foo') {
    sh('pwd')
  }
}
```
Wouldn't the shell step use the `FilePath` from the node rather than the adjusted one from `dir` like it would today?

dwnusbaum:
Also looks like we need to bump to plugin-pom 3.43 to fix the Javadoc failures.

jglick:
> bump to plugin-pom 3.43

Sure, I can do a round of updates.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: No, but only because I also patched the `dir` step. See b0c0fa6a2a695a076377843cab8ddac8c765f25f for discussion.

dwnusbaum: The `dir` and `ws` steps are distinct though, right? The following test fails for me with your changes, but if I revert src/main it passes:

```
diff --git a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
index 2c68d2d..78c9be4 100644
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -1166,6 +1166,20 @@ public class ExecutorStepTest {
         });
     }
 
+    @Test public void dirSmokes() {
+        story.then(r -> {
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p1");
+            p.setDefinition(new CpsFlowDefinition(
+                    "node {\n" +
+                    "  dir('custom') {\n" +
+                    "    sh('echo \"$(pwd) is script working dir\"')\n" +
+                    "  }\n" +
+            WorkflowRun b = r.buildAndAssertSuccess(p);
+            r.assertLogContains("/custom is script working dir", b);
+        });
+    }
+
```

jglick: Ah sorry, yes that needs to be patched as well, in a separate PR. I will prepare it.

dwnusbaum: I guess the bigger question is whether this matters for other types. I feel like `FilePath` is a bit special, and dynamic contexts for things like `TaskListenerDecorator` will avoid similar problems by being naturally additive with `TaskListenerDecorator.merge`, but it does seem a little concerning.

jglick: > `FilePath` is a bit special

In that all uses of it as a contextual object would suffer from the same issue reported for `node` in JENKINS-41854.

> dynamic contexts for things like `TaskListenerDecorator` will avoid similar problems by being naturally additive with `TaskListenerDecorator.merge`

Not automatically. [Discussion](https://github.com/jenkinsci/workflow-support-plugin/blob/79b6e6a9b308da2cec0040a735826d3709df24c9/src/test/java/org/jenkinsci/plugins/workflow/support/DynamicContextTest.java#L109-L209)

> it does seem a little concerning

Agreed. I was hoping to keep this change relatively self-contained but I can also spend some time looking into whether explicit support in `workflow-cps` could improve the semantics.

dwnusbaum: Is there any reason we can't just patch `DefaultStepContext` directly with something like this for `FilePath`? Will `get(Node.class)` fail if the `Computer` is no longer actually connected to the agent?

```
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java b/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
index 4ccdbcc..243f1ee 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/DefaultStepContext.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.workflow.support;
 
 import hudson.EnvVars;
+import hudson.FilePath;
 import hudson.Launcher;
 import hudson.LauncherDecorator;
 import hudson.console.ConsoleLogFilter;
@@ -38,6 +39,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
 import org.jenkinsci.plugins.workflow.flow.GraphListener;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
@@ -73,6 +75,16 @@ public abstract class DefaultStepContext extends StepContext {
             return key.cast(EnvironmentExpander.getEffectiveEnvironment(customEnvironment, (EnvVars) value, get(EnvironmentExpander.class), this, taskListener));
         } else if (key == Launcher.class) {
             return key.cast(makeLauncher((Launcher) value));
+        } else if (key == FilePath.class) {
+            FilePath fp = castOrNull(FilePath.class, value);
+            if (fp != null) {
+                Node n = get(Node.class);
+                if (n != null) {
+                    return castOrNull(key, FilePathUtils.find(n.getNodeName(), fp.getRemote()));
+                }
+            }
+            // Log warning?
+            return null;
         } else if (value != null) {
             return value;
         } else if (key == TaskListener.class) {
```

jglick: The above assumes that the contextual `Node` is the owner of the contextual `FilePath`. Generally that will be true, but it is not something I would want to rely on.

I think what you meant is something more like

```java
// …as before
if (fp != null) {
    VirtualChannel vch = fp.getChannel();
    if (vch instanceof Channel && ((Channel) vch).isClosingOrClosed()) {
        String node = FilePathUtils.getNodeNameOrNull(fp);
        if (node != null) {
            FilePath fp2 = FilePathUtils.find(node, fp.getRemote());
            if (fp2 != null) {
                return fp2;
            }
        }
    }
}
```

That would likely suffice to solve JENKINS-41854, at the expense of being hackish. I would prefer to have a general solution to the issue of injecting contextual objects or dynamically refreshing them.

Another consideration is that replacing `FilePathPickle` by `FilePathDynamicContext.FilePathRepresentation` in `program.dat` is a step away from relying on pickles, and thus the close coupling between restart semantics and Groovy engine internals. For example, a program cannot resume running until its `FilePathPickle` has been resolved, so in a program like

```groovy
parallel linux: {
  node('linux') {
    timeout(60) {
      sh 'make world'
    }
  }
}, win: {
  node('win') {
    timeout(60) {
      bat 'make world'
    }
  }
}
```

if Jenkins is restarted in the middle of both builds, and the Windows agent fails to reconnect, the build will hang indefinitely and you will not even be able to see if the Linux build has completed. Using `DynamicContext` only for `FilePath` will not solve this (there are other pickles which will hang here too), but it is a start.

dwnusbaum: > Another consideration is that replacing `FilePathPickle` by `FilePathDynamicContext.FilePathRepresentation` in `program.dat`...

Interesting, I missed that this means `FilePathPickle` wouldn't be used for new builds (unless some other plugin is adding a FilePath directly as a context variable). For builds across the restart, everything should be ok because `FilePathDynamicContext` won't find a `FilePathRepresentation` and will return null, and so `StepContext.doGet` will be used, right?

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/FilePathPickle.java
```diff
@@ -51,6 +51,9 @@ private FilePathPickle(FilePath v) {
             protected FilePath tryResolve() {
                 return FilePathUtils.find(slave, path);
             }
+            @Override protected FlowExecutionOwner getOwner() {
```
dwnusbaum: What is this for?

jglick: This was actually introduced for use from `PickleDynamicContext`, which was removed as of 5a7a0b29cad877d510074bc6c7d6f0a532214ce1, but it is independently useful—means that if a program _is_ still using `FilePathPickle` despite this PR + https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/86, and it is taking a while to resume, the build log will indicate why. Previously only `ExecutorPickle` implemented this. Generally pickles should implement this method if possible.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: > For builds across the restart

You mean, when updating plugins. Yes that is how it should work—`program.dat` will still contain a `FilePathPickle`, `FilePathDynamicContext.get` will return `null`, and `DefaultStepContext.get` will go back to checking `CpsStepContext.doGet(FilePath.class)`. Of course in this case JENKINS-41854 will still be observable:

```groovy
node {
  sh 'sleep 1h' // restarting during this
  sleep 9999 // disconnect agent during this
  sh ':' // would fail due to invalid Channel loaded during FilePathPickle.rehydrate
}
```

jglick:
On hold as in https://github.com/jenkinsci/workflow-support-plugin/pull/95#issuecomment-497823610.

dwnusbaum:
@jglick If someone updates to the new version of workflow-durable-task-step without also updating workflow-cps, things will totally break, since there won't be a `FilePath` in the context and `FilePathDynamicContext` won't actually be used, right? 

I was thinking we might be able to fix this by adding a `FilePath` in addition to `FilePathRepresentation`, but over in workflow-cps actual context variables take precedence over a `DynamicContext` at the same scope, so that would mean `FilePathRepresentation` would never be used.

Does this seem like something we should look into, or do you think a warning in the release notes is enough?

jglick:
I think it just needs to be release-noted. The Jenkins update center has no general way of dealing with this. Advertisement for Beekeeper I suppose.

PR #102, opened 2019-04-01
[JENKINS-50504] Demonstrate effectiveness of core fix in functional test

jglick:
Functional test for https://github.com/jenkinsci/jenkins/pull/3955. ~Subsumes https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/101.~

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.31</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.173</jenkins.version>
+        <jenkins.version>2.176</jenkins.version>
         <java.level>8</java.level>
         <workflow-step-api-plugin.version>2.20-rc522.e7bd30e9c518</workflow-step-api-plugin.version> <!-- TODO https://github.com/jenkinsci/workflow-step-api-plugin/pull/42 -->
         <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.32</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.150.1</jenkins.version>
+        <jenkins.version>2.176</jenkins.version>
```
jglick: The next LTS baseline.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -419,9 +419,7 @@ public void evaluate() throws Throwable {
                     Thread.sleep(100);
                 }
                 LOGGER.info("agent back online");
-                /* TODO JENKINS-50504 still fails:
                 assertWorkspaceLocked(computer, workspacePath);
```
dwnusbaum: It's not totally clear to me what the purpose of the PR is. Did this just fail intermittently before https://github.com/jenkinsci/jenkins/pull/3955? I was expecting to see some existing test become unignored or for a new test to be added.

jglick: The part of the test uncommented here fails consistently without the core patch.

dwnusbaum: Oh, duh, I was reading the comments as line comments rather than as the start and end of a block comment... Thanks for clearing that up!

PR #104, opened 2019-04-29
[JENKINS-49707] (part): Fail the build if an agent has been removed

jglick:
Improved behavior for one subcase of [JENKINS-49707](https://issues.jenkins-ci.org/browse/JENKINS-49707), that a disconnected agent is actually removed from Jenkins configuration for whatever reason. We assume that it is not going to be readded (under the same name), so there is no point in waiting for it to reconnect. ~The one exception that I know of is that a Swarm agent [will remove itself upon disconnect](https://github.com/jenkinsci/swarm-plugin/blob/796a04914e986f0fe6ab720e2cc95169831251dc/plugin/src/main/java/hudson/plugins/swarm/SwarmSlave.java#L89) but [might readd itself later](https://github.com/jenkinsci/swarm-plugin/blob/796a04914e986f0fe6ab720e2cc95169831251dc/plugin/src/main/java/hudson/plugins/swarm/PluginImpl.java#L219), so this might break cross-disconnect durability of Pipeline builds for Swarm, but that seems like a far less likely case than typical `Cloud`s.~

#47 and #48 do something similar across Jenkins restarts. I think the `TIMEOUT_WAITING_FOR_NODE_MILLIS` grace period there is irrelevant here, since that would be done before the program even resumes.

- [x] decide about `FAILURE` vs. `ABORTED` status
- [ ] ~test on EC2 with spot instances~ did not manage to set up the `ec2` plugin successfully
- [x] test against `kubernetes` plugin with a killed pod
- [x] consider implementing instead in `ExecutorStepExecution` (would work not just for `DurableTaskStep`)

@schottsfired

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -340,7 +340,7 @@ public abstract class DurableTaskStep extends Step {
                     // (Note that a Computer may be missing because a Node is offline,
                     // and conversely after removing a Node its Computer may remain for a while.
                     // Therefore we only fail here if _both_ are absent.)
-                    // ExecutorStepExecution.NodeListener will normally do this first, so this is a fallback.
+                    // ExecutorStepExecution.RemovedNodeListener will normally do this first, so this is a fallback.
                     Jenkins j = Jenkins.getInstanceOrNull();
                     if (ExecutorStepExecution.RemovedNodeCause.ENABLED && !node.isEmpty() && j != null && j.getNode(node) == null) {
                         if (removedNodeDiscovered == 0) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -58,6 +58,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import jenkins.model.Jenkins;
 import jenkins.util.Timer;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
@@ -321,6 +322,15 @@ public abstract class DurableTaskStep extends Step {
             if (ws == null) {
                 ws = FilePathUtils.find(node, remote);
                 if (ws == null) {
+                    // Part of JENKINS-49707: check whether an agent has been removed.
+                    // (Note that a Computer may be missing because a Node is offline,
+                    // and conversely after removing a Node its Computer may remain for a while.
+                    // Therefore we only fail here if _both_ are absent.)
+                    Jenkins j = Jenkins.getInstanceOrNull();
+                    if (!node.isEmpty() && j != null && j.getNode(node) == null) {
+                        // TODO or FlowInterruptedException?
+                        throw new AbortException("Agent " + node + " was removed");
+                    }
                     LOGGER.log(Level.FINE, "Jenkins is not running, no such node {0}, or it is offline", node);
                     return null;
                 }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -334,7 +334,7 @@ public abstract class DurableTaskStep extends Step {
                     // Therefore we only fail here if _both_ are absent.)
                     // ExecutorStepExecution.NodeListener will normally do this first, so this is a fallback.
                     Jenkins j = Jenkins.getInstanceOrNull();
-                    if (!node.isEmpty() && j != null && j.getNode(node) == null) {
+                    if (ExecutorStepExecution.RemovedNodeCause.ENABLED && !node.isEmpty() && j != null && j.getNode(node) == null) {
                         if (removedNodeDiscovered == 0) {
                             LOGGER.fine(() -> "discovered that " + node + " has been removed");
                             removedNodeDiscovered = System.nanoTime();
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -347,7 +347,7 @@ public abstract class DurableTaskStep extends Step {
                             LOGGER.fine(() -> "discovered that " + node + " has been removed");
                             removedNodeDiscovered = System.nanoTime();
                             return null;
-                        } else if (System.nanoTime() < removedNodeDiscovered + TimeUnit.MILLISECONDS.toNanos(ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS)) {
+                        } else if (System.nanoTime() - removedNodeDiscovered < TimeUnit.MILLISECONDS.toNanos(ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS)) {
                             LOGGER.fine(() -> "rediscovering that " + node + " has been removed");
                             return null;
                         } else {
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -34,6 +34,7 @@ import hudson.Launcher;
 import hudson.Util;
 import hudson.init.Terminator;
 import hudson.model.Node;
+import hudson.model.Result;
 import hudson.model.TaskListener;
 import hudson.remoting.Channel;
 import hudson.remoting.ChannelClosedException;
@@ -69,12 +70,14 @@ import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.graph.FlowNode;
 import org.jenkinsci.plugins.workflow.steps.AbstractStepExecutionImpl;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.Step;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepDescriptor;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
 import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
 import org.jenkinsci.plugins.workflow.support.concurrent.WithThreadName;
+import org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.DoNotUse;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -318,7 +321,7 @@ public abstract class DurableTaskStep extends Step {
             return false;
         }
 
-        private @CheckForNull FilePath getWorkspace() throws AbortException {
+        private @CheckForNull FilePath getWorkspace() throws IOException, InterruptedException {
             if (ws == null) {
                 ws = FilePathUtils.find(node, remote);
                 if (ws == null) {
@@ -326,10 +329,10 @@ public abstract class DurableTaskStep extends Step {
                     // (Note that a Computer may be missing because a Node is offline,
                     // and conversely after removing a Node its Computer may remain for a while.
                     // Therefore we only fail here if _both_ are absent.)
+                    // ExecutorStepExecution.NodeListener will normally do this first, so this is a fallback.
                     Jenkins j = Jenkins.getInstanceOrNull();
                     if (!node.isEmpty() && j != null && j.getNode(node) == null) {
-                        // TODO or FlowInterruptedException?
-                        throw new AbortException("Agent " + node + " was removed");
+                        throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
                     }
                     LOGGER.log(Level.FINE, "Jenkins is not running, no such node {0}, or it is offline", node);
                     return null;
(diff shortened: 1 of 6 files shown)


basil:
> this might break cross-disconnect durability of Pipeline builds for Swarm, but that seems like a far less likely case than typical Clouds

I rely on cross-disconnect durability of Pipeline builds for Swarm across my entire Jenkins fleet, so this would be a serious regression from my point of view. Would you be willing to run `PipelineJobTest#buildShellScriptAfterRestart` from jenkinsci/swarm-plugin#81 with these changes to ensure this workflow keeps working?

jglick:
> Would you be willing to run `PipelineJobTest#buildShellScriptAfterRestart` from [jenkinsci/swarm-plugin#81](https://github.com/jenkinsci/swarm-plugin/pull/81) with these changes

Sure, though I doubt that would reproduce any problem anyway, as this patch is about the case that a client is disconnected and removed while the Jenkins master is still running. To test _this_ patch you would need to kill the client and then have it restart, in a single “story”. Reconnection of a Swarm client across restarts is what #47/#48 should have been affecting; in that case there was a grace period, needed since `SwarmSlave` is an `EphemeralNode` and so we actually have to wait for it to be reattached.

----

@basil [this code](https://github.com/jenkinsci/swarm-plugin/blob/796a04914e986f0fe6ab720e2cc95169831251dc/plugin/src/main/java/hudson/plugins/swarm/PluginImpl.java#L213-L220) looks wrong to me. It could all be replaced by simply

```java
jenkins.addNode(slave);
```

since that method already replaces any existing node of the same name, without firing the deleted event. (Currently it always fires the created event, even when it ought to be firing updated; if I get a moment I will try to patch that.)

jglick:
> Currently it always fires the created event, even when it ought to be firing updated

https://github.com/jenkinsci/jenkins/pull/4004

basil:
> Sure, though I doubt that would reproduce any problem anyway, as this patch is about the case that a client is disconnected and removed while the Jenkins master is still running. To test _this_ patch you would need to kill the client and then have it restart, in a single “story”.

My latest update to jenkinsci/swarm-plugin#81 adds a new test for this case in `PipelineJobTest#buildShellScriptAfterDisconnect`.

basil:
> @basil [this code](https://github.com/jenkinsci/swarm-plugin/blob/796a04914e986f0fe6ab720e2cc95169831251dc/plugin/src/main/java/hudson/plugins/swarm/PluginImpl.java#L213-L220) looks wrong to me. It could all be replaced by simply
> 
> ```java
> jenkins.addNode(slave);
> ```
> 
> since that method already replaces any existing node of the same name, without firing the deleted event.

That change makes sense to me and my new integration tests still pass with it, so I opened jenkinsci/swarm-plugin#82.

jglick:
Even with https://github.com/jenkinsci/swarm-plugin/pull/82, `buildShellScriptAfterDisconnect` will fail due to the `removeNode` call from `SwarmSlave.SELF_CLEANUP_LAUNCHER.afterDisconnect`. I will check if a grace period helps.

jglick:
`hudson.plugins.swarm.PipelineJobTest` succeeds after

```diff
diff --git a/plugin/pom.xml b/plugin/pom.xml
index bfda46a..1a913a9 100644
--- a/plugin/pom.xml
+++ b/plugin/pom.xml
@@ -77,7 +77,7 @@
       </plugins>
     </build>
     <properties>
-        <workflow-support-plugin.version>2.13</workflow-support-plugin.version>
+        <workflow-support-plugin.version>3.3</workflow-support-plugin.version>
     </properties>
     <dependencyManagement>
         <dependencies>
@@ -90,19 +90,19 @@
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>scm-api</artifactId>
-                <version>1.3</version>
+                <version>2.2.6</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>structs</artifactId>
-                <version>1.5</version>
+                <version>1.18</version>
                 <scope>test</scope>
             </dependency>
             <dependency>
                 <groupId>org.jenkins-ci.plugins</groupId>
                 <artifactId>script-security</artifactId>
-                <version>1.24</version>
+                <version>1.58</version>
                 <scope>test</scope>
             </dependency>
         </dependencies>
@@ -111,7 +111,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.12</version>
+            <version>2.33</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -123,13 +123,13 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-cps</artifactId>
-            <version>2.28</version>
+            <version>2.70</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-durable-task-step</artifactId>
-            <version>2.4</version>
+            <version>2.32-rc917.b70f8edaf896</version> <!-- https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/104 -->
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -141,7 +141,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-step-api</artifactId>
-            <version>2.7</version>
+            <version>2.20</version>
             <scope>test</scope>
         </dependency>
         <dependency>
diff --git a/pom.xml b/pom.xml
index d5fbe91..4ce3480 100644
--- a/pom.xml
+++ b/pom.xml
@@ -24,7 +24,7 @@
     <version>3.18-SNAPSHOT</version>
 
     <properties>
-        <jenkins.version>2.60.3</jenkins.version>
+        <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
     </properties>
 
```

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -325,13 +332,34 @@ private static synchronized ScheduledThreadPoolExecutor threadPool() {
...
+                    // and conversely after removing a Node its Computer may remain for a while.
+                    // Therefore we only fail here if _both_ are absent.)
+                    // ExecutorStepExecution.NodeListener will normally do this first, so this is a fallback.
+                    Jenkins j = Jenkins.getInstanceOrNull();
+                    if (ExecutorStepExecution.RemovedNodeCause.ENABLED && !node.isEmpty() && j != null && j.getNode(node) == null) {
+                        if (removedNodeDiscovered == 0) {
+                            LOGGER.fine(() -> "discovered that " + node + " has been removed");
+                            removedNodeDiscovered = System.nanoTime();
+                            return null;
+                        } else if (System.nanoTime() < removedNodeDiscovered + TimeUnit.MILLISECONDS.toNanos(ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS)) {
```
dwnusbaum: Do we need to switch to the following as in the `System.nanoTime()` JavaDoc to avoid potential overflow issues (or just switch to `System.currentTimeMillis()` if nanosecond precision is not important)?

```suggestion
                        } else if (System.nanoTime() - removedNodeDiscovered < TimeUnit.MILLISECONDS.toNanos(ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS)) {
```

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -325,13 +332,34 @@ private static synchronized ScheduledThreadPoolExecutor threadPool() {
...
+                            LOGGER.fine(() -> "rediscovering that " + node + " has been removed");
+                            return null;
+                        } else {
+                            LOGGER.fine(() -> node + " has been removed for a while, assuming it is not coming back");
+                            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
```
dwnusbaum: Is `ABORTED` really the result we want in this case? My initial thought was that `FAILURE` would be more appropriate. Thoughts?

Looks like at least one reason for `ABORTED` is that over in `RemovedNodeListener` you call `BodyExecution.cancel`, and `CpsBodyExecution` always throws a `FlowInterruptedException` with result `ABORTED` in that method, so it makes sense to be consistent.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -325,13 +332,34 @@ private static synchronized ScheduledThreadPoolExecutor threadPool() {
...
-        private @CheckForNull FilePath getWorkspace() throws AbortException {
+        private @CheckForNull FilePath getWorkspace() throws IOException, InterruptedException {
             if (ws == null) {
                 ws = FilePathUtils.find(node, remote);
                 if (ws == null) {
+                    // Part of JENKINS-49707: check whether an agent has been removed.
+                    // (Note that a Computer may be missing because a Node is offline,
+                    // and conversely after removing a Node its Computer may remain for a while.
+                    // Therefore we only fail here if _both_ are absent.)
+                    // ExecutorStepExecution.NodeListener will normally do this first, so this is a fallback.
```
dwnusbaum:
```suggestion
                    // ExecutorStepExecution.RemovedNodeListener will normally do this first, so this is a fallback.
```

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -325,13 +332,34 @@ private static synchronized ScheduledThreadPoolExecutor threadPool() {
...
 
-        private @CheckForNull FilePath getWorkspace() throws AbortException {
+        private @CheckForNull FilePath getWorkspace() throws IOException, InterruptedException {
             if (ws == null) {
                 ws = FilePathUtils.find(node, remote);
                 if (ws == null) {
+                    // Part of JENKINS-49707: check whether an agent has been removed.
+                    // (Note that a Computer may be missing because a Node is offline,
+                    // and conversely after removing a Node its Computer may remain for a while.
+                    // Therefore we only fail here if _both_ are absent.)
```
dwnusbaum: Is that really the case here? It looks like we only check the Node, not the Computer.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -278,6 +326,17 @@ public void stop(Throwable cause) throws Exception {
          */
         private String cookie;
 
+        /**
+         * Needed for {@link BodyExecution#cancel}.
+         * {@code transient} because we cannot save a {@link BodyExecution} in {@link PlaceholderTask}:
+         * {@link ExecutorPickle} is written to the stream first, which holds a {@link PlaceholderTask},
+         * and the {@link BodyExecution} holds {@link PlaceholderTask.Callback} whose {@link WorkspaceList.Lease}
+         * is not processed by {@link WorkspaceListLeasePickle} since pickles are not recursive.
+         * So we make a best effort and only try to cancel a body within the current session.
```
dwnusbaum: So IIUC, if you restart Jenkins while a Pipeline is executing inside of a `node` block, and then remove the Node from Jenkins, that's how you get to the "Agent foo was deleted, but do not have a node body to cancel" branch in RemovedNodeListener? Or `session` here refers to something else?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: `currentTimeMillis` is inappropriate for deltas since it could drift with clock changes like NTP does.

The overflow issue only matters if Jenkins stays running for at least 292 years without a restart but OK!

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: Generally `FAILURE` is used for cases where something about the build itself broke. The typical scenario here is that the build started running, something like a spot instance got killed, and so the build could not complete due to external causes. Prior to this patch (plus downstream where appropriate) the build would just hang until manually `ABORTED`; now the abort is automatic.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Your understanding is correct. The patch to `DurableTaskStep` should still suffice to abort the build if it was inside `sh` when the node was removed.

If you remove the node _during_ the restart, a different bit of code in `ExecutorPickle` (existing long before this patch) kicks in and aborts the build before it even resumes the CPS VM thread.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: `FilePathUtils.find` above already checked nullness of `Jenkins.getComputer`.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
dwnusbaum: > currentTimeMillis is inappropriate for deltas since it could drift with clock changes like NTP does

Ah I didn't realize `System.nanoTime` was divorced from real-world time and only intended for elapsed time (a bit of selective Javadoc reading on my part).

slaughter550:
@jglick Quick question, we came across this and I was wondering how to set `org.jenkinsci.plugins.workflow.support.pickles.ExecutorPickle.timeOutForNodeMillis` correctly? We tried setting it on jenkins master with an arg on Jenkins startup with `-D` but it doesn't appear to be inheriting correctly. I'm new to Jenkins so apologies if it's a stupid question. Thanks!

jglick:
`-D` is used for this sort of thing. Maybe try the users’ mailing list?

slaughter550:
For anyone who finds this later, its `org.jenkinsci.plugins.workflow.support.pickles.ExecutorPickle.timeoutForNodeMillis` where timeout is all lower case. The changelog [here](https://wiki.jenkins.io/pages/viewpage.action?pageId=96797030#PipelineNodesandProcessesPlugin-2.14(Aug23,2017)) is incorrect

jglick:
@slaughter550 you can file a PR for https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/master/CHANGELOG.md#232 as needed though I see no mention of the property to be misspelled.

slaughter550:
@jglick PS submitted, its not misspelled, its miscaptatolized. https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/142

PR #106, opened 2019-06-10
Added EXECUTOR_COUNT to environment variables

ArturHarasimiuk:
It may be reasonable to check executor count of currently used master/slave agent. This PR exposes this as environment variable.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
@@ -100,6 +100,29 @@ public class EnvWorkflowTest {
         r.assertLogContains("My number on a slave is 0", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
+    /**
+     * Verifies if EXECUTOR_COUNT environemn variable is set
+     */
+    @Test public void isExecutorCountAvailable() throws Exception {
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -100,6 +100,29 @@
         r.assertLogContains("My number on a slave is 0", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
     }
 
+    /**
+     * Verifies if EXECUTOR_COUNT environemn variable is set
```
dwnusbaum:
```suggestion
     * Verifies if EXECUTOR_COUNT environment variable is set
```

ArturHarasimiuk:
I don't have corresponding Jenkins JIRa, at least I didn't look for this. I have hetero
Having agents with different HW configuration and different number of executors. My sample usages are:
1. when monitoring ccache usage during build, it is not reliable with more than one executor because of concurrent accesses from parallel builds and race condition during ccache stats cleaning and read.
2. tune behavior to reduce system load by reducing parallelism

jglick:
`EXECUTOR_NUMBER` is defined in Jenkins core and implemented there for freestyle builds. I do not think Pipeline builds should have this feature unless it is also added for freestyle.

PR #109, opened 2019-06-25
Update durable-task dependency and add notes indicating potential zombies

car-roll:
Test PR showing that unit tests pass when using the durable task plugin version from [PR-92](https://github.com/jenkinsci/durable-task-plugin/pull/92)


car-roll:
Updated to using incemental from [durable-task 106](https://github.com/jenkinsci/durable-task-plugin/pull/106)

dwnusbaum:
Would be good to go ahead and update the help for `sh` here as discussed in https://github.com/jenkinsci/durable-task-plugin/pull/106#discussion_r310727556.

dwnusbaum:
It would be great to update this with the latest version of the upstream PR. You can `mvn deploy` the upstream plugin to avoid issues with the CI infra for now.

car-roll:
@dwnusbaum okay will do. Thought I had to wait until the CI had to upload to incrementals :p

src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep/help-script.html
```diff
@@ -9,4 +9,10 @@
     Otherwise the system default shell will be run, using the <code>-xe</code> flags
     (you can specify <code>set +e</code> and/or <code>set +x</code> to disable those).
     </p>
+    <p>
+    NOTE: The script is launched and monitored using a binary with the prefix <code>durable-task-monitor-</code>.
+          Because the binary is launched as a background process, please ensure that the target script environment
+          is able to reap orphaned processes (i.e. zombie-reaping). For example, in docker, you may use the
```
dwnusbaum:
```suggestion
          is able to reap orphaned processes (i.e. zombie-reaping). For example, in Docker, you may use the
```

Maybe worth linking to https://docs.docker.com/engine/reference/run/?#specify-an-init-process (you can use `a` tags here).

src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep/help-script.html
```diff
@@ -9,4 +9,10 @@
...
+          <code>--init</code> flag, while in kubernetes, you may leverage shared namespaces and use the pause container.
```
dwnusbaum:
```suggestion
          <code>--init</code> flag, while in Kubernetes, you may leverage shared namespaces and use the pause container.
```

If there are any official docs on this, it could be good to link to them. Personally I don't know how to do this in Kubernetes.

car-roll: Will do. I actually should also verify this, at the very least, locally via `minikube`

src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep/help-script.html
```diff
@@ -9,4 +9,14 @@
...
     </p>
+    <p>
+    NOTE: The script is launched and monitored using a binary with the prefix <code>durable-task-monitor-</code>.
+          Because the binary is launched as a background process, please ensure that the target script environment
+          is able to reap orphaned processes (i.e. zombie-reaping). For example, in Docker, you may use the
+          <a href="https://docs.docker.com/engine/reference/run/?#specify-an-init-process"><code>--init</code> flag</a>,
+          while in Kubernetes, you may leverage
+          <a href="https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/"> shared process namespace</a>.
+          With shared process namespaces, the pod sandbox will be assigned PID 1 and assume
+          <a href="https://github.com/kubernetes/kubernetes/blob/master/build/pause/pause.c#L37">zombie-reaping responsibilities</a>.
```
car-roll: Maybe that's too specific? I just couldn't find any official documentation on the zombie reaping for some reason. I found PRs and discussion and blog posts, but nothing on kubernetes.io. Maybe i'm searching for the wrong keywords?

dwnusbaum: Yeah, I wouldn't link directly to source code. I think the link to shared process namespaces is good enough. Any idea of how common this kind of setting is, or if there are any alternatives? It looks like it has some drawbacks in terms of security, so it seems less than ideal if that is the only way to get things working in Kubernetes.

car-roll: It's actually enabled by default (as of `v1.15`)

car-roll: The only other way i'm aware of in Kubernetes, would be to then roll your own `init` process for each container

dwnusbaum: If I understand correctly, the ability to use the feature is what is enabled by default, but you have to explicitly use it by adding `shareProcessNamespace: true` to the Pod spec. I think disabling the feature makes it so that `shareProcessNamespace` is not a valid field for Pod specs.

car-roll: Looks like the feature has been discussed for a few years now, but only got introduced as a beta in its current incarnation about a year ago. That's about the only insight I can give about how common of a use case it is.

car-roll:
Ran simple test on kubernetes with the following [Jenkinsfile](https://gist.github.com/car-roll/4471dc9231abeb1db1271ad349706bcb).
In the [subsequent output](https://gist.github.com/car-roll/0dc1f124090ebee569cce4af17db6d1b), you can see that a zombie process exists in the first stage where there is no Shared Process Namespace versus the second stage where there Shared Process Namespace is enabled.

dwnusbaum:
> hudson.plugins.git.GitException: org.eclipse.jgit.api.errors.TransportException: Premature EOF

Rebuilding.

src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStep/help-script.html
```diff
@@ -9,4 +9,13 @@
     Otherwise the system default shell will be run, using the <code>-xe</code> flags
     (you can specify <code>set +e</code> and/or <code>set +x</code> to disable those).
     </p>
+    <p>
+    NOTE: The script is launched and monitored using a binary with the prefix <code>durable-task-monitor-</code>.
```
dwnusbaum: nit: maybe clarify that the binary is only used on some OSs

jglick:
Is this about https://issues.jenkins.io/browse/JENKINS-58656? Does `tini` suffice rather than `shareProcessNamespace: true`?

car-roll:
@jglick Yes, if you specify `tini` as en entrypoint that would work too.
fyi: https://github.com/kubernetes/kubernetes/issues/84210

PR #112, opened 2019-07-23
Final line of shell output without newline lost when using LineTransformationOutputStream-based decorators

jglick:
Fixing the test from #103, strengthening it in the process. Taking on some urgency since after @Vlatombe’s https://github.com/jenkinsci/kubernetes-plugin/pull/552, this problem would affect any `sh` step in a Kubernetes agent which prints output without a trailing newline (no longer limited to that minority of builds which deliberately use secrets).

- [X] tested fix
- [X] verification that `kubernetes` test workarounds can be removed: https://github.com/jenkinsci/kubernetes-plugin/pull/554
- [X] verification that partial lines are sanely handled in `pipeline-cloudwatch-logs`: https://github.com/jenkinsci/pipeline-cloudwatch-logs-plugin/pull/58

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,6 +65,7 @@
         <revision>2.33</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.176.1</jenkins.version>
+        <jenkins-test-harness.version>2.54-rc1207.a9cdd7c1192a</jenkins-test-harness.version> <!-- TODO https://github.com/jenkinsci/jenkins-test-harness/pull/149 -->
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
@@ -79,7 +80,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.29</version>
+            <version>1.31-rc427.7bfff1482201</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/102 -->
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -132,7 +132,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials-binding</artifactId>
-            <version>1.16</version>
+            <version>1.17</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
         <revision>2.33</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.176.1</jenkins.version>
-        <jenkins-test-harness.version>2.54-rc1207.a9cdd7c1192a</jenkins-test-harness.version> <!-- TODO https://github.com/jenkinsci/jenkins-test-harness/pull/149 -->
+        <jenkins-test-harness.version>2.54</jenkins-test-harness.version> <!-- TODO pending plugin-pom update -->
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
@@ -80,7 +80,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.31-rc427.7bfff1482201</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/102 -->
+            <version>1.29</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -132,7 +132,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>credentials-binding</artifactId>
-            <version>1.16</version>
+            <version>1.17</version>
```
jglick: To pick up https://github.com/jenkinsci/credentials-binding-plugin/pull/51. This seems to be the reason that, in watch mode, the test was previously failing (output was getting lost) even in situations that did not involve a missing final newline.

pom.xml
```diff
@@ -65,6 +65,7 @@
         <revision>2.33</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.176.1</jenkins.version>
+        <jenkins-test-harness.version>2.54-rc1207.a9cdd7c1192a</jenkins-test-harness.version> <!-- TODO https://github.com/jenkinsci/jenkins-test-harness/pull/149 -->
```
jglick: https://github.com/jenkinsci/jenkins-test-harness/pull/149

pom.xml
```diff
@@ -79,7 +80,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.29</version>
+            <version>1.31-rc427.7bfff1482201</version> <!-- TODO https://github.com/jenkinsci/durable-task-plugin/pull/102 -->
```
jglick: https://github.com/jenkinsci/durable-task-plugin/pull/102

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -417,6 +426,50 @@ private void getWorkspaceProblem(Exception x) {
...
+            }
+
+            // Similar to DecoratedTaskListener:
+            @Override public synchronized PrintStream getLogger() {
+                if (logger == null) {
+                    LOGGER.fine("creating filtered stream");
+                    OutputStream base = delegate.getLogger();
+                    OutputStream filtered = new FilterOutputStream(base) {
+                        boolean nl = true; // empty string does not need a newline
+                        @Override public void write(int b) throws IOException {
```
dwnusbaum: Is this in a context where performance matters? If so, maybe good to override the other `write` overloads, especially since I think just looking at the last byte would still work fine for bulk writes.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -599,6 +652,7 @@ private void handleExit(int exitCode, OutputSupplier output) throws IOException,
                     getContext().onFailure(new AbortException("script returned exit code " + exitCode));
                 }
             }
+            listener().getLogger().close();
```
dwnusbaum: So this call will always be on `NewlineSafeTaskListener`, right? Is this why we don't want to call `super.close()` in `NewlineSafeTaskListener.$NewlineSafeTaskListener.close()`? Ditto for the call below in `exited`. Also, if we are calling `close` here now, why didn't we call `flush` here earlier? Should we have been doing that?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: True, overloading `write(byte[], int, int)` could be slightly faster.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick: > So this call will always be on `NewlineSafeTaskListener`, right?

Right.

> Is this why we don't want to call `super.close()` in `NewlineSafeTaskListener.$NewlineSafeTaskListener.close()`?

Yes.

> if we are calling `close` here now, why didn't we call `flush` here earlier?

There is no particular need to flush until the end of the step (and `close` flushes).

dwnusbaum: > There is no particular need to flush until the end of the step (and close flushes).

My question was unclear, but my confusion is about why we weren't calling `flush` before this PR, since it looks like this is the end of the step.

jglick: Oh, I see! We _were_ calling `flush` before, in watch mode, in `exited`. In non-watch mode, it did not really matter since the log messages were coming from the master, and `DefaultStepContext.getListener` always `close`s the per-step listener when the step finishes.

jglick:
> I am struggling to keep track of when to flush or not, what overrides of `close` also close their delegates, and what is getting remoted or not in Pipeline logging in general

So am I!

pom.xml
```diff
@@ -65,6 +65,7 @@
         <revision>2.33</revision>
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.176.1</jenkins.version>
+        <jenkins-test-harness.version>2.54</jenkins-test-harness.version> <!-- TODO pending plugin-pom update -->
```
jglick: https://github.com/jenkinsci/plugin-pom/pull/221

PR #114, opened 2019-07-30
Trivial code cleanup for tests

basil:
A series of trivial code cleanup changes for the tests in this repository. Most of these changes remove usages of deprecated functionality or otherwise fix IDE warnings. I kept each commit separate. Each commit message should be self-explanatory, but if anything is unclear I am happy to provide further explanation. Also, feel free to revert any specific cleanup commits that are unwanted.


src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
```diff
@@ -126,52 +114,6 @@
         });
     }
 
-    /** Ephemeral and deleted when it disconnects */
-    static class EphemeralDumbAgent extends Slave implements EphemeralNode {
```
dwnusbaum: I was confused about this earlier as well, looks like it was added in #47, but then the test using it was deleted in #48, but the class stayed around unused.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -182,12 +183,12 @@
...
-                File f1 = new File(story.j.jenkins.getRootDir(), "test.txt");
-                String fullPathToTestFile = f1.getAbsolutePath();
+                Path f1 = story.j.jenkins.getRootDir().toPath().resolve("test.txt");
+                String fullPathToTestFile = f1.toAbsolutePath().toString();
                 // Escape any \ in the source so that the script is valid
                 fullPathToTestFile = fullPathToTestFile.replace("\\", "\\\\");
-                // Ensure deleted
-                f1.delete();
+                // Ensure file does not exist
+                assertFalse(Files.exists(f1));
```
dwnusbaum: Did the old behavior matter if this plugin's test were being run in the same workspace more than once? I guess you could use `Files.deleteIfExists(f1)` to preserve the old behavior, but I'm not sure what the original concern was.

src/test/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickleTest.java
basil: Indeed. Fun fact: #48 was reverting #47 because it broke a production use case for me!

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
basil: I too was having a hard time understanding the original behavior. Your explanation is as good as any, so we might as well preserve the existing functionality. I switched to `Files.deleteIfExists` and updated the comment to reflect your explanation.

dwnusbaum:
> java.nio.channels.ClosedChannelException

The Windows agent was broken, rebuilding.

dwnusbaum:
Another `java.nio.channels.ClosedChannelException` on the Windows agent, rebuilding.

basil:
> Looks good, I'll merge once the CI build passes (ping me if it completes and it looks like I forgot about it smile).

I think this should be good to go now.

PR #115, opened 2019-08-12
[JENKINS-58900] Improve error messages for agent disconnections inside of node

dwnusbaum:
See [JENKINS-58900](https://issues.jenkins-ci.org/browse/JENKINS-58900).

The problem is that you get a `MissingContextVariableException` where you would expect to get some kind of connection-related exception (seen on some builds of Jenkins core on ci.jenkins.io where the Windows agent apparently disconnected but the error in the logs was a `MissingContextVariableException`). I think this is a side effect of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/101 ([JENKINS-41854](https://issues.jenkins-ci.org/browse/JENKINS-41854)), but I haven't investigated enough to know for sure. If so, maybe fixable by throwing some kind of exception instead of just logging here: https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/5840b563c6c0cd9b7f721bcaff55e4b005be234d/src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java#L60

CC @oleg-nenashev

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
@@ -26,10 +26,12 @@ package org.jenkinsci.plugins.workflow.support.steps;
 
 import hudson.Extension;
 import hudson.FilePath;
+import hudson.model.Computer;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.BodyInvoker;
 import org.jenkinsci.plugins.workflow.steps.DynamicContext;
@@ -57,7 +59,14 @@ import org.jenkinsci.plugins.workflow.support.pickles.FilePathPickle;
         if (f != null) {
             LOGGER.log(Level.FINE, "serving {0}:{1}", new Object[] {r.slave, r.path});
         } else {
-            LOGGER.log(Level.FINE, "failing to serve {0}:{1}", new Object[] {r.slave, r.path});
+            IOException e = new IOException("Unable to create live FilePath for " + r.slave);
+            Computer c = Jenkins.get().getComputer(r.slave);
+            if (c != null) {
+                for (Computer.TerminationRequest tr : c.getTerminatedBy()) {
+                    e.addSuppressed(tr);
+                }
+            }
+            throw e;
         }
         return f;
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
@@ -27,6 +27,8 @@ package org.jenkinsci.plugins.workflow.support.steps;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Computer;
+import hudson.model.TaskListener;
+import hudson.slaves.OfflineCause;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.logging.Level;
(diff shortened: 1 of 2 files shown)


dwnusbaum:
> java.nio.file.FileSystemException: /home/jenkins/.jenkins/cache/jars/3F: No space left on device

I filed https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/116 to hopefully make the build a bit more robust.

dwnusbaum:
> win2012-4df5c0 was marked offline: Connection was broken: java.util.concurrent.TimeoutException: Ping started at 1565643975690 hasn't completed by 1565644215691

😩

src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
```diff
@@ -57,7 +59,14 @@
         if (f != null) {
             LOGGER.log(Level.FINE, "serving {0}:{1}", new Object[] {r.slave, r.path});
         } else {
-            LOGGER.log(Level.FINE, "failing to serve {0}:{1}", new Object[] {r.slave, r.path});
```
dwnusbaum: Not obvious to me if we were returning `null` here for any reason in particular. `ExecutorStepTest.contextualizeFreshFilePathAfterAgentReconnection` passes with the change, not sure if there are any other scenarios related to `FilePathDynamicContext` where the new behavior would be problematic.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1161,6 +1161,33 @@ String n(Run<?, ?> b, String label) {
...
+            p.setDefinition(new CpsFlowDefinition(
+                    "node ('" + agent.getNodeName() + "') {\n" +
+                    "  def isUnix = isUnix()\n" + // Only call `isUnix()` before the agent goes offline to avoid additional log warnings.
+                    "  isUnix ? sh('echo hello') : bat('echo hello')\n" +
+                    "  semaphore('wait')\n" +
+                    "  isUnix ? sh('echo world') : bat('echo hello')\n" +
+                    "}", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", b);
+            agent.toComputer().disconnect(new OfflineCause.UserCause(User.getUnknown(), "going offline"));
```
jglick: Should we also see `going offline` in the log?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1161,6 +1161,33 @@ String n(Run<?, ?> b, String label) {
...
+    @Test public void nodeDisconnectMissingContextVariableException() {
+        story.then(r -> {
+            DumbSlave agent = r.createOnlineSlave();
+            WorkflowJob p = r.createProject(WorkflowJob.class);
+            p.setDefinition(new CpsFlowDefinition(
+                    "node ('" + agent.getNodeName() + "') {\n" +
+                    "  def isUnix = isUnix()\n" + // Only call `isUnix()` before the agent goes offline to avoid additional log warnings.
+                    "  isUnix ? sh('echo hello') : bat('echo hello')\n" +
+                    "  semaphore('wait')\n" +
+                    "  isUnix ? sh('echo world') : bat('echo hello')\n" +
```
jglick:
```suggestion
                    "  isUnix ? sh('echo world') : bat('echo world')\n" +
```

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
dwnusbaum: Yeah, that would be good, so I added code to print the `OfflineCause` in 8d51aac (copied from `ExecutorStepExecution`) and added an assertion in the test.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
```diff
@@ -66,6 +68,13 @@
                     e.addSuppressed(tr);
                 }
             }
+            TaskListener listener = context.get(TaskListener.class);
+            if (listener != null) {
+                OfflineCause oc = c.getOfflineCause();
+                if (oc != null) {
+                    listener.getLogger().println(c.getDisplayName() + " was marked offline: " + oc);
```
jglick: Note that this assumes `OfflineCause.toString` has been meaningfully overridden. That is definitely true of `SimpleOfflineCause`, and seems to be done for other subtypes as well, though really it ought to have been defined as `abstract` to ensure that it is implemented.

PR #117, opened 2019-08-19
Release node body reference when finishing a PlaceholderTask

basil:
### Problem

After upgrading `workflow-durable-task-step` from 2.31 to 2.32 in `workflow-cps-global-lib-plugin`, `LibraryMemoryTest#loaderReleased` starts failing with the following stack trace:

```
[ERROR] loaderReleased(org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest)  Time elapsed: 1.561 s  <<< ERROR!
java.lang.NullPointerException
	at org.netbeans.insane.live.Path.getField(Path.java:146)
	at org.netbeans.insane.live.Path.toString(Path.java:140)
	at org.netbeans.insane.live.Path.toString(Path.java:138)
	at java.lang.String.valueOf(String.java:2994)
	at java.lang.StringBuilder.append(StringBuilder.java:131)
	at java.util.AbstractMap.toString(AbstractMap.java:559)
	at java.lang.String.valueOf(String.java:2994)
	at java.lang.StringBuilder.append(StringBuilder.java:131)
	at org.jvnet.hudson.test.MemoryAssert.assertGC(MemoryAssert.java:187)
	at org.jenkinsci.plugins.workflow.libs.LibraryMemoryTest.loaderReleased(LibraryMemoryTest.java:85)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.rules.ExternalResource$1.evaluate(ExternalResource.java:48)
	at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:599)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:298)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:292)
```

### Evaluation

Bisection revealed that the problem does not occur with `workflow-durable-task-step` 2.31 (see jenkinsci/workflow-cps-global-lib-plugin#80) but does occur with `workflow-durable-task-step` 2.32 (see jenkinsci/workflow-cps-global-lib-plugin#81). Looking at the changes [in the compare view](https://github.com/jenkinsci/workflow-durable-task-step-plugin/compare/workflow-durable-task-step-2.31...workflow-durable-task-step-2.32), I noticed two new fields: `DurableTaskStep#removedNodeDiscovered` and `PlaceholderTask#body`, both added in jenkinsci/workflow-durable-task-step-plugin#104. `DurableTaskStep#removedNodeDiscovered` is a `long` and is unlikely to be the leak, so I hypothesized that `PlaceholderTask#body` was the leak. Sure enough, when I changed `body = context.newBodyInvoker().[...]` back to `context.newBodyInvoker().[...]` in `PlaceholderExecutable#run` (the only place where `PlaceholderExecutable#body` was ever set) and re-ran `LibraryMemoryTest#loaderReleased`, it passed. This confirmed that the proximate source of the problem was `PlaceholderTask#body`.

While I don't know enough about this plugin to be completely sure, based on the above analysis I strongly suspect that the cause of the problem is that `PlaceholderTask#body` is being leaked, exposing the `workflow-cps-global-lib-plugin` test failure in `LibraryMemoryTest#loaderReleased`.

### Solution

Assuming that the above evaluation is correct, the solution would be to release `PlaceholderExecutable#body` when it is no longer needed. Without knowing very much about this plugin, `PlaceholderTask#finish` seemed like as good a place as any to do this. Much to my surprise, `workflow-cps-global-lib-plugin`'s `LibraryMemoryTest#loaderReleased` test started passing again once I made this change.

If I am misunderstanding the cause of the problem or if there is a better solution, please feel free to let me know.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -277,12 +277,13 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         } catch (Exception x) {
                             LOGGER.log(Level.WARNING, null, x);
                         }
-                        if (task.body == null) {
+                        BodyExecution body = task.body != null ? task.body.get() : null;
+                        if (body == null) {
                             listener.getLogger().println("Agent " + node.getNodeName() + " was deleted, but do not have a node body to cancel");
                             continue;
                         }
                         listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        task.body.get().cancel(new RemovedNodeCause());
+                        body.cancel(new RemovedNodeCause());
                     }
                 }
             }, ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -837,7 +837,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                             flowNode.addAction(new WorkspaceActionImpl(workspace, flowNode));
                         }
                         listener.getLogger().println("Running on " + ModelHyperlinkNote.encodeTo(node) + " in " + workspace);
-                        body = new WeakReference<BodyExecution>(context.newBodyInvoker()
+                        body = new WeakReference<>(context.newBodyInvoker()
                                 .withContexts(exec, computer, env,
                                     FilePathDynamicContext.createContextualObject(workspace))
                                 .withCallback(new Callback(cookie, lease))
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -34,6 +34,7 @@ import hudson.slaves.WorkspaceList;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.io.Serializable;
+import java.lang.ref.WeakReference;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -281,7 +282,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                             continue;
                         }
                         listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        task.body.cancel(new RemovedNodeCause());
+                        task.body.get().cancel(new RemovedNodeCause());
                     }
                 }
             }, ExecutorPickle.TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -725,6 +725,16 @@ private static void finish(@CheckForNull final String cookie) {
                     // JENKINS-30759: finished before asynch execution was even scheduled
                     return;
                 }
+                final Executor executor = execution.getExecutor();
+                if (executor != null) {
+                    final Queue.Executable exec = executor.getCurrentExecutable();
+                    if (exec instanceof PlaceholderTask.PlaceholderExecutable) {
+                        final PlaceholderExecutable placeholderExec = (PlaceholderExecutable) exec;
+                        if (placeholderExec.getParent().body != null) {
```
jglick: Unnecessary check.

basil:
> Seems fine, though it would be easier to make `body` be a `WeakReference<BodyExecution>` I suppose.

Sure, I reimplemented the change with `body` as a `WeakReference<BodyExecution>`. Feel free to let me know if this isn't what you had in mind.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -281,7 +282,7 @@ public void stop(Throwable cause) throws Exception {
                             continue;
                         }
                         listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        task.body.cancel(new RemovedNodeCause());
+                        task.body.get().cancel(new RemovedNodeCause());
```
basil: Does the result of `WeakReference#get()` need to be surrounded with a `null` check? After reading the Javadoc more closely, I think it does. But I think I will wait for confirmation from a reviewer before I go through another testing cycle.

dwnusbaum: I'm not sure, it might be non-null in this case but it's not obvious to me that the `WeakReference` could never be `null` here, so I would add `BodyExecution body = task.body == null ? null : task.body.get();`  on line 280 before the `task.body == null` check, and then change the check to `body == null`, and use `body` here at which point it would be guaranteed non-null.

basil: I'm not so sure either. `WeakReference#get` does state that the result can be "`null` if this reference object has been cleared". Presumably whatever is actually running the body has a reference to it until it stops running (which is why this implementation strategy is safe in the first place), but I'm not familiar enough with Pipeline internals to confirm this. In any case, I went ahead and added the `null` check as you suggested. I'm normally against overly-defensive `null` checks, but since I'm not the primary maintainer of this code it seemed prudent to avoid the risk of any potential regressions.

basil:
> You checked that the `WeakReference` rewrite solves the test failure?

Yes, of course. But don't take my word for it, see the results for yourself in jenkinsci/workflow-cps-global-lib-plugin#82.

PR #118, opened 2019-09-23
Fix race condition in ShellStepTest.abort

slonopotamus:
The problem is that build cancellation is async and build can touch the file
between the moment when test read file modification time and build process was killed.

Note that this commit doesn't fully fix test on Windows due to [JENKINS-59152](https://issues.jenkins-ci.org/browse/JENKINS-59152).


slonopotamus:
jenkinsci/jenkins#4216 is accompanying this change in order to make `ShellStepTest.abort` stable on Windows.

dwnusbaum:
A slightly modified form of this was added in #137 to fix the test, thanks!

PR #121, opened 2019-10-07
fix label in step generator

jetersen:
when using the generator it would return `powershell label: '', script: 'test'`


casz:
Believe test needs to be updated I just never got around to this PR again, thanks for the reminder.

PR #125, opened 2019-12-06
Remove notice of security advisories

jglick:
Unnecessary here, as https://plugins.jenkins.io/workflow-durable-task-step displays these automatically. I guess this got blindly copied from the wiki, where it is not markup but a macro.


dwnusbaum:
Thanks, I already filed https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/124 for this, will go ahead and merge that.

PR #126, opened 2019-12-06
[JENKINS-60389] Optimization for getOwnerTask

jglick:
[JENKINS-60389](https://issues.jenkins-ci.org/browse/JENKINS-60389)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -59,8 +59,6 @@ import jenkins.security.QueueItemAuthenticatorProvider;
 import jenkins.util.Timer;
 import org.acegisecurity.AccessDeniedException;
 import org.acegisecurity.Authentication;
-import org.acegisecurity.context.SecurityContext;
-import org.acegisecurity.context.SecurityContextHolder;
 import org.jenkinsci.plugins.durabletask.executors.ContinuableExecutable;
 import org.jenkinsci.plugins.durabletask.executors.ContinuedTask;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -433,6 +431,13 @@ public String getCookie() {
...
 
         @Override public Queue.Task getOwnerTask() {
+            Jenkins j = Jenkins.getInstanceOrNull();
+            if (j != null && runId != null) { // JENKINS-60389 shortcut
+                Job<?, ?> job = j.getItemByFullName(runId.substring(0, runId.lastIndexOf('#')), Job.class);
+                if (job instanceof Queue.Task) {
+                    return (Queue.Task) job;
+                }
+            }
             Run<?,?> r = runForDisplay();
```
dwnusbaum: nit: It looks like `runId` should be non-null since at least version 2.0, so maybe fine to `return this` here and drop the old code entirely (unless there is some way to rename/move a Job while a build is running (renaming should be disallowed by `Job.checkRename`) that would cause the context-based lookup to succeed even though the `runId`-based lookup did not, in which case you could switch to `run()` to avoid calling `Run.fromExternalizableId`).

jglick: > It looks like `runId` should be non-null since at least version 2.0, so maybe fine to `return this` here and drop the old code entirely

We probably could. I preferred to make this patch be conservative and just address the performance issue without making a potentially breaking change. Cleanup such as you describe could be a follow-up with more careful review of the impact.

> renaming should be disallowed by `Job.checkRename`

Yes, it should be, so I do not believe there is a valid reason to _not_ trust `runId` when it is defined.

Note that there are also other places where `run` or `runForDisplay` is being called which _could_ be optimized via `runId` but which I did not bother to touch, in the absence of a particular reason. I believe the issue in this case was specific to startup, when a queue task was being processed before build loading was complete, and that may not be relevant for other methods. Just speculating—I do not know how to reproduce the issue.

PR #128, opened 2020-01-03
[JENKINS-60634] Bind $WORKSPACE_TMP from the node & ws steps

jglick:
Like https://github.com/jenkinsci/jenkins/pull/4414 but for Pipeline builds.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -1197,6 +1197,17 @@ public class ExecutorStepTest {
         });
     }
 
+    @Test
+    @Issue("JENKINS-60634")
+    public void tempDirVariable() throws Exception {
+        story.then(r -> {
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1197,6 +1197,17 @@ String n(Run<?, ?> b, String label) {
         });
     }
 
+    @Test
+    @Issue("JENKINS-60634")
+    public void tempDirVariable() throws Exception {
+        story.then(r -> {
+            WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("node {if (isUnix()) {sh 'set -u && touch \"$WORKSPACE_TMP/x\"'} else {bat(/echo ok > \"%WORKSPACE_TMP%\\x\"/)}}", true));
```
jglick: Note that unlike in the core (freestyle) test, there is no need to `mkdir $WORKSPACE_TMP` here: the control directory for `durable-task` is inside `p@tmp` already.

PR #132, opened 2020-03-25
Prevent lease leak when ws or node is called with no body

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Currently calling the ws step with no body will leak a lease of the workspace requested for in that step, as the step will fail due to not having a body for bodyinvoker. Check that there is a body and if no body is supplied fail early.

```
node {
    ws('/some/path') // /some/path leaks and subsequent runs will not be able to use it without @2 being appended
}
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -101,6 +101,9 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
      */
     @Override
     public boolean start() throws Exception {
+        if(!getContext().hasBody()) {
+            throw new AbortException("The node step must be called with a body");
+        }
         final PlaceholderTask task = new PlaceholderTask(getContext(), step.getLabel());
         Queue.WaitingItem waitingItem = Queue.getInstance().schedule2(task, 0).getCreateItem();
         if (waitingItem == null) {
@@ -793,9 +796,6 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                     Executor exec = Executor.currentExecutor();
                     if (exec == null) {
                         throw new IllegalStateException("running task without associated executor thread");
-                    }
-                    if (!context.hasBody()) {
-                        throw new Exception("node must be called with a body");
                     }
                     computer = exec.getOwner();
                     // Set up context for other steps inside this one.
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -794,6 +794,9 @@ public void run() {
                     if (exec == null) {
                         throw new IllegalStateException("running task without associated executor thread");
                     }
+                    if (!context.hasBody()) {
+                        throw new Exception("node must be called with a body");
```
dwnusbaum: Would it be better to make this the first line of `ExecutorStepExecution.start` so we don't even get to the point of adding a `PlaceholderTask` to the queue? Other than that, I might change the exception type since we don't care about a stack trace, and tweak the message:

```suggestion
                        throw new AbortException("The node step must be called with a body"); // Requires an import of hudson.model.AbortException
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -35,6 +35,9 @@ public boolean start() throws Exception {
         if (!(job instanceof TopLevelItem)) {
             throw new Exception(job + " must be a top-level job");
         }
+        if (!getContext().hasBody()) {
+            throw new Exception("ws must be called with a body");
```
dwnusbaum: I'd use `AbortException` since we don't care about the stack trace, and tweak the message slightly:

```suggestion
            throw new AbortException("The ws step must be called with a body"); // Requires an import of hudson.model.AbortException
```

res0nance:
Made the appropriate changes, Yeah moving it to `start()` is a better idea.

jglick:
I think this ought to be handled in `workflow-cps` by throwing a syntax error.

bitwiseman:
@jglick 
Do you mean we should not merge this change? Or are you making a non-blocking comment?

dwnusbaum (changes requested):
Thinking about Jesse's comment, I agree that it doesn't make sense to handle this here since we'd want to do the same thing for nearly every block-scoped step.

@car-roll is making changes to the code related to running steps with/without bodies in https://github.com/jenkinsci/workflow-cps-plugin/pull/370 anyway, so I think it would make sense to go ahead and add a more robust check for steps without bodies there (perhaps with a special case for `stage`).

jglick:
Yeah `StageStep` is the tricky case because @kohsuke decided to use the same step name for a completely different syntax and implementation, so we had to relax the original check that `takesImplicitBlockArgument` is interpreted strictly. Perhaps this should have been done with another method in `StepDescriptor`, like a three-value `enum`.

res0nance:
I think its perfectly reasonable to handle it there, the reason this existed was there did not seem to be anything else in the works

car-roll:
Closing this PR now that https://github.com/jenkinsci/workflow-cps-plugin/pull/370 has been merged

PR #134, opened 2020-03-31
Use CauseOfBlockage.print rather than Queue.Item.why

jglick:
https://github.com/jenkinsci/jenkins/blob/cc27faa48d8c5a0cf421a5d648231b506391c2eb/core/src/main/java/hudson/model/Queue.java#L2267-L2270 is `final` so we may as well allow it to use rich text. Useful in conjunction with https://github.com/jenkinsci/jenkins/pull/4616.


jglick:
ping @dwn

PR #135, opened 2020-05-11
[JENKINS-62014] Add support for build step environment filters

daniel-beck:
See [JENKINS-62014](https://issues.jenkins-ci.org/browse/JENKINS-62014).

This adds support for global build step environment filters to durable task based pipeline steps.

Downstream from https://github.com/jenkinsci/jenkins/pull/4683 and https://github.com/jenkinsci/durable-task-plugin/pull/124.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.36</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.236-SNAPSHOT</jenkins.version>
+        <jenkins.version>2.248</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.35-SNAPSHOT</version>
+            <version>1.35-rc463.dbcd6297773f</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.54</version>
+        <version>4.1</version>
         <relativePath />
     </parent>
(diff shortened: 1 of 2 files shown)


jeffret-b:
There's a build failure issue with dependencies.

jeffret-b:
This looks good, though I'm not an expert in this area. There's a build error with dependencies so I'm not able to try it out or go further with it.

pom.xml
```diff
@@ -79,7 +79,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>durable-task</artifactId>
-            <version>1.33</version>
+            <version>1.35-SNAPSHOT</version>
```
jeffret-b, line 82:
```suggestion
            <version>1.35-rc462.28a634e20bc4</version>
```

jeffret-b:
I tested basic operations with a few different configurations. Everything worked as expected.

The console log messages are a little weird, but it looks like they don't originate here.

This PR could use an improved description of what it adds and how it needs to be tested.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -302,6 +303,7 @@ private static synchronized ScheduledThreadPoolExecutor threadPool() {
             ws = context.get(FilePath.class);
             node = FilePathUtils.getNodeName(ws);
             DurableTask durableTask = step.task();
+            durableTask.setRun(context.get(Run.class));
```
dwnusbaum, line 306: @daniel-beck It looks like a few lines down in this method, you would have access to the `launcher`, so you could call `launcher.prepareFilterRules(context.get(Run.class), durableTask);` (not sure what context makes the most sense, could pass `durableTask` or `this` I guess...) without needing to make any changes in durable-task (other than the addition of `implements EnvVarsFilterableBuilder` on `FileMonitoringTask`, but I am not totally clear on whether it needs to be there or could be here on `DurableTaskStep`).

Any particular reason you wanted to make changes to each kind of task over in `durable-task`? I only glanced at the Jenkins core PR so may very well be missing something obvious.

daniel-beck: > other than the addition of `implements EnvVarsFilterableBuilder` on `FileMonitoringTask`, but I am not totally clear on whether it needs to be there or could be here on `DurableTaskStep`

I want to allow global rules to determine what they do based on the kind of build step; so I cannot use the same generic type from this plugin for all of {Batch; Shell; Powershell}.

daniel-beck:
Superseded by #137.

PR #136, opened 2020-06-11
[JENKINS-26097] Adjust label validation and auto-completion

Zastai:
Uses the API being introduced for JENKINS-26097 (https://github.com/jenkinsci/jenkins/pull/4774)

(Also squishes a spotbugs whine.)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>3.54</version>
+        <version>4.2</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.36</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.1</jenkins.version>
+        <jenkins.version>2.241-rc30038.0428cd872667</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.36</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.241-rc30038.0428cd872667</jenkins.version> <!-- https://github.com/jenkinsci/jenkins/pull/4774 -->
+        <jenkins.version>2.243</jenkins.version>
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <workflow-step-api-plugin.version>2.20</workflow-step-api-plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -100,7 +100,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-job</artifactId>
-            <version>2.31</version>
+            <version>2.33</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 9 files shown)


Zastai:
At first glance, the test failures don't seem to come from these changes. Perhaps compatibility issue with newer Jenkins?

jglick:
> the test failures don't seem to come from these changes

No. I would think https://github.com/jenkinsci/jenkins-test-harness/pull/210 would have solved all these, as in https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/116#discussion_r436741358, but you _did_ update the parent here. @dwnusbaum any idea offhand? I can dig into it.

dwnusbaum:
> any idea offhand?

Best guess is that it has to do with issues fixed in [workflow-job 2.33+](https://github.com/jenkinsci/workflow-job-plugin/blob/master/CHANGELOG.md#233). This plugin is still using 2.31, so I would try updating dependencies to see if that makes a difference. @Zastai If you are interested, you can try pulling the latest version of https://github.com/jenkinsci/bom into this PR to update everything, or we can handle it in a separate PR.

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.36</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.1</jenkins.version>
+        <jenkins.version>2.241-rc30038.0428cd872667</jenkins.version>
```
jglick, line 67:
```suggestion
        <jenkins.version>2.241-rc30038.0428cd872667</jenkins.version> <!-- https://github.com/jenkinsci/jenkins/pull/4774 -->
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -838,7 +838,9 @@ public void run() {
                         // Cf. AbstractBuild.getEnvironment:
                         env.put("WORKSPACE", workspace.getRemote());
                         if (workspace.getParent() != null) {
-                            env.put("WORKSPACE_TMP", WorkspaceList.tempDir(workspace).getRemote()); // JENKINS-60634
+                            FilePath tmp = WorkspaceList.tempDir(workspace);
+                            if (tmp != null)
+                                env.put("WORKSPACE_TMP", tmp.getRemote()); // JENKINS-60634
```
jglick, line 843:
```suggestion
                            if (tmp != null) {
                                env.put("WORKSPACE_TMP", tmp.getRemote()); // JENKINS-60634
                            }
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepExecution.java
```diff
@@ -63,7 +63,9 @@ public boolean start() throws Exception {
         Map<String, String> env = new HashMap<>();
         env.put("WORKSPACE", workspace.getRemote());
         if (workspace.getParent() != null) {
-            env.put("WORKSPACE_TMP", WorkspaceList.tempDir(workspace).getRemote());
+            FilePath tmp = WorkspaceList.tempDir(workspace);
+            if (tmp != null)
+                env.put("WORKSPACE_TMP", tmp.getRemote());
```
jglick, line 68:
```suggestion
            if (tmp != null) {
                env.put("WORKSPACE_TMP", tmp.getRemote());
            }
```

Zastai:
Looks like the bump to workflow-job 2.33 did it; made the tests pass for me.

I did have to add an explicit dependency on junit 1.20 or the InjectedTest (which I think comes from the plugin parent) failed (because it seems to use matrix-project 1.14, which wants junit 1.20, and apparently 1.6 was being pulled in).

dwnusbaum:
I think the `ExecutorStepTest.buildShellScriptAcrossRestart` failure is similar to flakes investigated in #139, so I pushed a fix into https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/121 and will rebuild this once that PR build passes and is merged.

PR #137, opened 2020-07-27
[JENKINS-62014] Add support for global build step env var filters

daniel-beck:
After feedback from @dwnusbaum in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/135#discussion_r459644019, this is now a much simpler implementation.

### Testing notes

Install https://plugins.jenkins.io/generic-environment-filters/ 1.1 or newer, and configure any of the global filters it provides in the global configuration.

* The easiest to test is "Set environment variable" as it just adds a new environment variable.
* The most comprehensive test case is "Filter environment variables..." as that allows the definition of job name exclusions, and can be configured to only apply to certain build steps. Both features work with this PR, although the latter is a bit messy because freestyle and pipeline steps have similar names, and show up in the same list here. That's just a https://plugins.jenkins.io/environment-filter-utils/ issue though that shouldn't matter here.

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,2 +1,3 @@
 -Pconsume-incrementals
 -Pmight-produce-incrementals
+-Duse-jenkins-bom
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,3 +1,2 @@
 -Pconsume-incrementals
 -Pmight-produce-incrementals
--Duse-jenkins-bom
--- a/pom.xml
+++ b/pom.xml
@@ -162,5 +162,11 @@
             <version>1.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>junit</artifactId>
+            <version>1.29</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -159,7 +159,7 @@
         <dependency>
             <groupId>io.jenkins.plugins</groupId>
             <artifactId>generic-environment-filters</artifactId>
-            <version>1.1</version>
+            <version>1.2</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.4</version>
+        <version>4.6</version>
         <relativePath />
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
@@ -84,7 +84,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
-            <version>2.33</version>
+            <version>2.40</version>
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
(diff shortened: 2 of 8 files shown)


daniel-beck:
> FTR the plugin "global-environment-filters" is not yet available in the update-center, only from https://repo.jenkins-ci.org/.

Sorry, I meant https://plugins.jenkins.io/generic-environment-filters/ , fixed the comment.

pom.xml
```diff
@@ -64,7 +64,7 @@
     <properties>
         <revision>2.36</revision>
         <changelist>-SNAPSHOT</changelist>
-        <jenkins.version>2.176.1</jenkins.version>
+        <jenkins.version>2.248</jenkins.version>
```
dwnusbaum, line 67: Hmm, I guess because of the new interface there is no way to get things working on older versions?

.mvn/maven.config
```diff
@@ -1,2 +1,3 @@
 -Pconsume-incrementals
 -Pmight-produce-incrementals
+-Duse-jenkins-bom
```
dwnusbaum, line 3: What is this for?

pom.xml
daniel-beck: Yes, it uses a new core type that was added in this core release. But the feature supported here is only available starting with this core release anyway, so that's not a problem.

Supporting older baselines in this plugin might require backports etc. though, so I can understand if you want to hold off the merge for a bit longer.

.mvn/maven.config
daniel-beck: It's a magic "fix dependency problems" switch 🤷 Maybe it's no longer needed, will check.

dwnusbaum: I checked the current version of `plugin-pom` and did not see it, so I was not sure what it was for specifically. Maybe it was required in the beta 4.x releases to enable the Jenkins BOM? I think I fixed some version conflicts with core dependencies by updating to 4.x versions of the parent POM on other plugins without needing to add this, although maybe that was something else.

dwnusbaum:
@daniel-beck This looks good to me, but I want to add a basic test to make sure we don't accidentally break this in the future, so I filed https://github.com/daniel-beck/workflow-durable-task-step-plugin/pull/1, could you please take a look at that and merge it into your branch and/or rework as desired?

daniel-beck:
There are several assertions for text in the build log that only appears in stderr:

- `org.jenkinsci.plugins.workflow.steps.durable_task.ShellStepTest#interruptingAbortsBuildEvenWithReturnStatus`
- `org.jenkinsci.plugins.workflow.support.steps.ExecutorStepTest#queueItemAction`
- `org.jenkinsci.plugins.workflow.support.steps.ExecutorStepTest#queueTaskVisibility`

@dwnusbaum Any idea whether that is expected, perhaps due to a dependency update? I don't want to waste time investigating if you know where this is coming from.

daniel-beck:
Weird, these tests pass locally on my Mac, with only the assertion in `ExecutorStepTest#265` failing.

dwnusbaum:
I'm not sure why the tests are failing, but I will try to look today.

dwnusbaum:
I can't reproduce any of the failures locally. My best guess is that they are related to flakiness with the behavior of `JenkinsRule.waitForCompletion` in relation to the last few lines of log output when Pipelines are finishing, and that they only show up in the CI environment because it's slower than our local machines or something. These issues have been fixed in recent versions of `workflow-cps`, `workflow-job`, and `jenkins-test-harness`, so I refiled this PR with another commit that uses the plugin BOM to update all plugin dependencies in case that helps, see https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/139.

dwnusbaum:
Ok, based on the investigation in #139, I've pushed all of the fixes into this PR. 🤞

PR #138, opened 2020-07-27
[JENKINS-41940] Add weight support for node step 

goganchic:
[JENKINS-41940](https://issues.jenkins.io/browse/JENKINS-41940)

Typical Jenkins setup consists of the main server with several worker nodes. Each node has several slots. By default one build occupies one slot. Some builds will wait in a queue if there are no more free slots available.

Each node has a limited amount of resources like CPU, RAM, or network. Different builds require different amount of these resources but each build occupies only one slot. That is why in some cases one node can be overloaded and another almost idle. Ability to set weight for each build as [heavy-job-plugin](https://github.com/jenkinsci/heavy-job-plugin) can solve this problem. Unfortunately heavy-job-plugin works only with Freestyle projects but not with Pipeline.

Recently I've created [pipeline-heavy-job-plugin](https://github.com/funbox/pipeline-heavy-job-plugin) but in [HOSTING-1005](https://issues.jenkins-ci.org/browse/HOSTING-1005) Jenkins contributors suggested me to modify existing plugin instead of creating new one.

This PR allows to use such syntax in Jenkinsfile:

```
node(weight: 2) {
  performSomeHeavyTask
}
```

It means that such build will occupy two slots instead of one. After this feature will be merged into workflow-durable-task-step-plugin it will be possible to add weight support into Pipeline Declarative Syntax in [pipeline-model-definition-plugin](https://github.com/jenkinsci/pipeline-model-definition-plugin).


goganchic:
@jglick would you please review my PR?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1244,4 +1244,47 @@ public String getShortDescription() {
...
+                b.getExecutor().interrupt();
+                story.j.assertBuildStatus(Result.ABORTED, story.j.waitForCompletion(b));
+            }
+        });
+    }
+
+    @Test public void heavyBuildWaitingInQueue() throws Exception {
+        story.addStep(new Statement() {
+            @Override public void evaluate() throws Throwable {
+                DumbSlave s = story.j.createOnlineSlave();
```
jglick, line 1277: Specify number of executors explicitly. The default changed to 1.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -426,7 +429,35 @@ public String getCookie() {
         }
 
         @Override public Collection<? extends SubTask> getSubTasks() {
-            return Collections.singleton(this);
+            List<SubTask> r = new ArrayList<SubTask>();
+            r.add(this);
+            for (int i = 1; i < weight; i++)
+                r.add(new SubTask() {
```
jglick, line 435: I do not think this will serialize correctly. `JenkinsSessionRule` must be used to verify behavior across restarts.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -997,6 +1028,33 @@ public boolean hasPermission(Permission permission) {
             }
         }
 
+        public static class ExecutableImpl implements Queue.Executable {
```
jglick, line 1031: Needs to be an `AsynchronousExecutable` I think, or whatever it is called.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
```diff
@@ -67,16 +68,29 @@
 
     private static final long serialVersionUID = 1L;
 
-    private final @CheckForNull String label;
+    private @CheckForNull String label;
```
jglick, line 71: revert
```suggestion
    private final @CheckForNull String label;
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
```diff
@@ -67,16 +68,29 @@
...
 
     @DataBoundConstructor public ExecutorStep(String label) {
         this.label = Util.fixEmptyAndTrim(label);
     }
-    
+
+    @DataBoundSetter public void setLabel(String label) {
+        this.label = Util.fixEmptyAndTrim(label);
+    }
+
```
jglick, line 81: Revert, you cannot mix `@DataBoundConstructor` and `@DataBoundSetter`.

```suggestion
```

If you want to set a weight but _not_ a label, you will need to

```groovy
node(label: null, weight: 2)
```

But in practice you would almost certainly want both.

goganchic: Would you please explain your point a bit more? Why I can't use `@DataBoundConstructor` and `@DataBoundSetter` at the same time? My idea is to preserve current behavior (`node('myLabel')`) but add an ability to set label with weight: `node(label: 'myLabel', weight: 2)` and weight only: `node(weight: 2)`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
goganchic: Unfortunately I don't understand the idea. Do you mean that ExecutableImpl should extend AsynchronousExecution? Why? Where can I get more info about this question?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
goganchic: Do you mean something like this https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/138/files#diff-649f51b85e49696fd7f4e4fb6ee16c61R1260-R1283?

goganchic:
> Probably incorrect; definitely undertested. This is a nontrivial change to one of the more subtle and error-prone parts of Pipeline. On the other hand, the risk of regression is low since the modified behavior is clearly self-contained.

Would you please tell me which tests should I add?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
jglick: Any given property is either mandatory (DBC) or optional (DBS). In the case of a single mandatory property, there is a special syntax whereby the name may be omitted, hence `node('label') {…}`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Something like that, using either `RestartableJenkinsRule` or its successor `JenkinsSessionRule`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: This is a complicated topic and goes to the root of how Pipeline builds survive Jenkins restarts and specifically how a `sh` step running inside a `node` block will resume. `ExecutorPickle` is used to save the `PlaceholderTask` representing the agent’s executor slot. `PlaceholderExecutable` does some setup when the `node` block is entered, then throws `RunningTask.execution` to represent the fact that it is running without literally consuming a Java thread; this is redone after a restart, hence the check for `cookie == null`.

jglick:
I suspect doing this correctly is a lot more difficult than the code in this PR would suggest.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep.java
goganchic: thanks!

goganchic:
> I suspect doing this correctly is a lot more difficult than the code in this PR would suggest.

@jglick would you please give me a piece of advice what source code should I explore to implement this feature correctly?

slonopotamus:
1. How does it behave if I have only one worker nodes with only one executor and run a task with weight=2?
2. How does it behave if I have a two workers with only one executor each and run a task with weight=2?
3. How does it behave if I have a two nodes with two executors (but one executor is busy while other one is free) each and run a task with weight=2?

goganchic:
>     1. How does it behave if I have only one worker node with only one executor and run a task with weight=2?

The task will never start, because you don't have enough resources to execute it.

>     2. How does it behave if I have two workers with only one executor each and run a task with weight=2?

The task will never start, because you don't have enough resources to execute it.

>     3. How does it behave if I have a two nodes with two executors (but one executor is busy while other one is free) each and run a task with weight=2?

The task will be waiting in a queue for two free slots on one node.

jglick:
> what source code should I explore

`ExecutorStepExecution.java` & `ExecutorPickle.java`. Unfortunately I think this system is not really documented at a high level.

Have not yet checked what actually happens after a restart here—whether the current simplistic implementation happens to work reliably due to some mechanism I have not seen explained, or whether it only works in the limited circumstances covered by tests.

jglick:
> Is there any possibility to set log level = FINE for Jenkins process started by test?

Yes, use `LoggerRule`.

goganchic:
@jglick as I can see, it's required to define which loggers I want to record. I don't know exactly which logger can give me some useful information. Is it possible to tell `LoggerRule` to show all possible logs with `FINEST` level?

jglick:
> Is it possible to tell `LoggerRule` to show all possible logs with `FINEST` level?

Well, `.record("")` but be prepared to get a new hard drive to capture all the output…

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
goganchic: fixed, I guess

goganchic:
@jglick I'd changed source code as you suggested: `PlaceholderSubTaskExecutable` is a `AsynchronousExecution` and tests pass on my localhost. Would you please take a look on the PR one more time?

goganchic:
@jglick @dwnusbaum @car-roll should I prepare this PR for review somehow?

jglick:
Not sure I have time to review it right now, but I am not a maintainer of this plugin any more either.

dwnusbaum:
@goganchic I will try to review this in the next few weeks. In the meantime it would be helpful if you ran a build of the plugin based on this PR on your instance to look for issues related to build resumption and termination when using the new `weight` parameter.

goganchic:
Hello!

@dwnusbaum any updates?

goganchic:
I'd checked build resumption and termination when using the new `weight` parameter on my local machine - for my tasks everything works fine.

oleg-nenashev:
CC @dwnusbaum @jenkinsci/code-reviewers

goganchic:
@dwnusbaum @jenkinsci/code-reviewers any news on this PR? Should I fix or improve something to make it possible to merge?

goganchic:
@dwnusbaum @oleg-nenashev I'll be happy if you give me some feedback. I'm waiting for it almost for 10 months and still want to merge this PR :) The feature from the PR looks useful. If merging is impossible, please tell me and I'll close the PR.

car-roll:
hi @goganchic , I'll be trying to take some time to go over this PR. I'm sorry for the delay, it's been on my todo list for some time but I've had a lot of various priorities pop up, both personal and work related. As busy as I am, I'm pretty sure that @dwnusbaum or @oleg-nenashev have even less time at the moment.

oleg-nenashev:
TBH I see little point in spending it untill the plugin maintainers can dedicate time to this plugin. I dedicated time to reviewing and testing other Pipeline PRs over the past years, but then they were stuck on merge/release I am not ready to become a co-maintainer at the moment. As a Jenkins board member. I am happy to work with the team on the Pipeline ecosystem sustainability plan.

goganchic:
@car-roll @oleg-nenashev any news about this PR?

servolkov:
well, it seems there is no plan to finish this one? It's sad, I thought the feature is basical one (in terms of load balance).

PR #133, opened 2020-03-25
Use non-deprecated lazily evaluated iterateEnclosingBlocks

res0nance:
Signed-off-by: Raihaan Shouhell <raihaan.shouhell@autodesk.com>

Replacing the deprecated FlowScanningUtils.
See:
https://github.com/jenkinsci/workflow-api-plugin/blob/f21a0577338eecb418d9f354de600a1a509e84f9/src/main/java/org/jenkinsci/plugins/workflow/graphanalysis/FlowScanningUtils.java#L113-L124

PR #140, opened 2020-08-15
Code cleanup

basil:
A series of code cleanup changes for this repository. Most of these changes remove usages of deprecated functionality or otherwise fix IDE warnings. I kept each commit separate. Each commit message should be self-explanatory, but if anything is unclear I am happy to provide further explanation. I suggest ignoring whitespace changes in the GitHub UI when reviewing 
f06bd78. Also, feel free to revert any specific cleanup commits that are unwanted.


dwnusbaum:
Two test failures, one I think is a legitimate flaky test, and I am investigating in #133. The other looks like a CI infra issue:

> java.lang.OutOfMemoryError: unable to create new native thread

Perhaps they are both caused by the infra issue, but I'm not sure.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -193,7 +191,7 @@ public void stop(Throwable cause) throws Exception {
             }
             Jenkins j = Jenkins.getInstanceOrNull();
             if (j != null) {
-                COMPUTERS: for (Computer c : j.getComputers()) {
```
dwnusbaum, line 196: Weird...

PR #144, opened 2020-11-16
[JENKINS-63486] Use ACL.SYSTEM in PlaceholderTask.getOwnerTask as before JENKINS-60389

dwnusbaum:
See [JENKINS-63486](https://issues.jenkins.io/browse/JENKINS-63486). Amends #126.

[JENKINS-63868](https://issues.jenkins.io/browse/JENKINS-63868) was fixed in recent versions of Jenkins core, but it will still log a warning every time this code throws an exception, so we should change this code anyway.

I added a regression test in 2c1da66, but I have not attempted to reproduce the problem or test the fix on a real Jenkins instance.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -431,9 +431,11 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         @Override public Queue.Task getOwnerTask() {
             Jenkins j = Jenkins.getInstanceOrNull();
             if (j != null && runId != null) { // JENKINS-60389 shortcut
-                Job<?, ?> job = j.getItemByFullName(runId.substring(0, runId.lastIndexOf('#')), Job.class);
-                if (job instanceof Queue.Task) {
-                    return (Queue.Task) job;
+                try (ACLContext context = ACL.as(ACL.SYSTEM)) {
+                    Job<?, ?> job = j.getItemByFullName(runId.substring(0, runId.lastIndexOf('#')), Job.class);
+                    if (job instanceof Queue.Task) {
+                        return (Queue.Task) job;
+                    }
                 }
             }
             Run<?,?> r = runForDisplay();
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -31,6 +31,7 @@ import hudson.Functions;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Item;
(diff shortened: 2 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -431,9 +431,11 @@ public String getCookie() {
...
-                if (job instanceof Queue.Task) {
-                    return (Queue.Task) job;
+                try (ACLContext context = ACL.as(ACL.SYSTEM)) {
+                    Job<?, ?> job = j.getItemByFullName(runId.substring(0, runId.lastIndexOf('#')), Job.class);
+                    if (job instanceof Queue.Task) {
+                        return (Queue.Task) job;
+                    }
                 }
             }
             Run<?,?> r = runForDisplay();
```
dwnusbaum, line 441: This is the fallback code that was used 100% of the time before #126, and it uses `ACL.as(ACL.SYSTEM)`, so using `ACL.as(ACL.SYSTEM)` in the new code seems fine.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1216,6 +1218,30 @@ public void tempDirVariable() {
...
+
+            r.jenkins.setSecurityRealm(r.createDummySecurityRealm());
+            r.jenkins.setAuthorizationStrategy(new MockAuthorizationStrategy()
+                    .grant(Job.DISCOVER).onFolders(f).toEveryone()
+                    .grant(Job.READ).onItems(p).toEveryone());
+            User alice = User.get("alice", true, Collections.emptyMap());
+
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            SemaphoreStep.waitForStart("wait/1", b);
+            for (Executor e : Jenkins.get().toComputer().getExecutors()) {
```
dwnusbaum, line 1237: There is probably a direct way to get the executor running the `node` step, but I could not think of it offhand.

car-roll (approved):

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1216,6 +1218,30 @@ public void tempDirVariable() {
...
+                try (ACLContext context = ACL.as(alice)) {
+                    e.hasStopPermission(); // Throws AccessDeniedException before JENKINS-63486.
```
dwnusbaum, line 1239: There might be a clearer way to test this, maybe just calling `e.getExecutable.getParent().getOwnerTask()` directly, although we'd need to make sure the executable was non-null.

MRamonLeon (approved):

jglick:
see https://github.com/jenkinsci/jenkins/pull/5013

PR #147, opened 2020-11-24
ExecutorStepExecution.queue_task_cancelled should yield ABORTED not FAILURE

jglick:
If you click the red **X** next to a build under the master node, or in the build log, the build is marked `ABORTED` like you would expect. But if you do the same on a heavyweight task on an executor, perhaps just because that is the most convenient place to find it, the build is marked `FAILURE` which seems inconsistent.


schottsfired (approved):
The new test passed:
image

One failed, but it was an existing failure.

timja:
Needs a test change here https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/master/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java#L702 to ABORTED

jglick:
Artifactory seems to be having difficulty today.

car-roll (approved):

PR #150, opened 2021-02-27
failing ExecutorStepTest in jenkins versions > 2.65

car-roll:
When running Jenkins versions > 2.65, the `ExecutorStepTest.reuseNodesWithSameLabelsInDifferentReorderedStages` test will fail.

This is due to changes in `LoadBalancer`  that were done in 2.66 here: https://github.com/jenkinsci/jenkins/pull/5028

Added a note to indicate this in the test code.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 buildPlugin(useAci: true, configurations: [
-    [platform: 'linux', jdk: '8'],
-    [platform: 'windows', jdk: '8'],
-    [platform: 'linux', jdk: '11']
+    [platform: 'linux', jdk: '8', jenkins: null],
+    [platform: 'windows', jdk: '8', jenkins: null],
+    [platform: 'linux', jdk: '11', jenkins: "2.277", javaLevel: 8]
 ])
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 buildPlugin(useAci: true, configurations: [
-    [platform: 'linux', jdk: '8', jenkins: null],
-    [platform: 'windows', jdk: '8', jenkins: null],
+    [platform: 'linux', jdk: '8'],
+    [platform: 'windows', jdk: '8'],
     [platform: 'linux', jdk: '11', jenkins: "2.277", javaLevel: 8]
 ])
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -862,6 +862,8 @@ public boolean apply(@Nullable FlowNode input) {
     @Issue("JENKINS-36547")
     @Test public void reuseNodesWithSameLabelsInDifferentReorderedStages() {
         story.then(r -> {
+            // Note: for Jenkins versions > 2.65, the number of agents must be increased to 5.
+            // This is due to changes in the Load Balancer (See JENKINS-60563).
             for (int i = 0; i < 3; ++i) {
```
dwnusbaum, line 867: You can do something like this so the test passes against all versions:

```suggestion
            // Note: for Jenkins versions > 2.265, the number of agents must be increased to 5.
            // This is due to changes in the Load Balancer (See JENKINS-60563).
            int totalAgents = Jenkins.getVersion().isNewerThan(new VersionNumber("2.265")) ? 5 : 3;
            for (int i = 0; i < totalAgents; ++i) {
```

You would also need to add an import for `hudson.util.VersionNumber` and I would update the Jenkinsfile to test against both the current baseline `jenkins: null` and the LTS candidate (`jenkins: 2.277`, and I would add this one to the `java: 11` branch rather than adding another parallel branch).

Jenkinsfile
```diff
@@ -1,5 +1,5 @@
 buildPlugin(useAci: true, configurations: [
-    [platform: 'linux', jdk: '8'],
-    [platform: 'windows', jdk: '8'],
-    [platform: 'linux', jdk: '11']
+    [platform: 'linux', jdk: '8', jenkins: null],
+    [platform: 'windows', jdk: '8', jenkins: null],
```
dwnusbaum, line 3: nit: `jenkins: null` is the default, so you can leave it out if desired:

```suggestion
    [platform: 'linux', jdk: '8'],
    [platform: 'windows', jdk: '8'],
```

dwnusbaum (approved):

PR #156, opened 2021-04-28
Add Bat Online Help

Queen-esther01:
I added some examples to the bat step online help in the node and processes plugin as well references to the syntax generator to encourage users to try out examples.


MarkEWaite:
@kwhetstone , @aHenryJard , and @stackscribe, would any of you be able to help me with this review.  I'm running behind schedule and could really use some help

PR #159, opened 2021-04-30
Add Return Stdout Help

Queen-esther01:
I added two examples to the returnStdout argument using declarative and scripted pipelines using the sh step and bat step alongside a link to the steps reference and syntax generator


MarkEWaite:
@kwhetstone , @aHenryJard , and @stackscribe, would any of you be able to help me with this review.  I'm running behind schedule and could really use some help

PR #162, opened 2021-05-23
Rename PowershellScriptStep to WindowsPowerShellScriptStep

Zastai:
Aside from capitalization (PowerShell), the current branding used by Microsoft is "Windows PowerShell" for version 5.1 (and earlier,
although those versions seem to be out of support) and just "PowerShell" for the multiplatform implementation (i.e. no more "Core").
I have not (yet) dropped the Core part from the other step class, because it is still in very common use (including in much of Microsoft's own materials).

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/WindowsPowerShellScriptStep.java
```diff
@@ -32,11 +32,11 @@
 /**
  * Asynchronous batch script execution.
  */
-public class PowershellScriptStep extends DurableTaskStep {
+public class WindowsPowerShellScriptStep extends DurableTaskStep {
```
dwnusbaum, line 35: Please don't change class names or move files (unless absolutely necessary) to avoid issues with backwards compatibility (for example, I found [this reference](https://github.com/jenkinsci/labelled-steps-plugin/blob/a7015e81cb60f0adb5baad170412c730bfe2577a/src/main/java/me/vickychijwani/jenkins/LabelledPowerShellStep.java#L70) to the existing class in a plugin). I would recommend reverting the moves/renames and leaving only the change in `getDisplayName`.

dwnusbaum (changes requested):
Thanks for the PR!

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/WindowsPowerShellScriptStep.java
Zastai: Fair enough. Only thought to do these branding fixes because I saw another rename come in when I updated my fork of this repo. But looking now, that was only a case fix in a resource folder name, not a class rename.

timja:
@car-roll may as well include this in release?

car-roll (approved):

PR #167, opened 2021-09-07
Make this test compatible with Jenkins 2.307+

daniel-beck:
https://www.jenkins.io/changelog/#v2.307 / https://github.com/jenkinsci/jenkins/pull/5425

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
@@ -46,14 +46,15 @@ public class EnvWorkflowTest {
      */
     @Test public void isNodeNameAvailable() throws Exception {
         r.createSlave("node-test", "unix fast", null);
+        String builtInNodeLabel = r.jenkins.getSelfLabel().getExpression(); // compatibility with 2.307+
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "workflow-test");
 
         p.setDefinition(new CpsFlowDefinition(
-            "node('master') {\n" +
+            "node('" + builtInNodeLabel + "') {\n" +
             "  echo \"My name on master is ${env.NODE_NAME} using labels ${env.NODE_LABELS}\"\n" +
             "}\n",
             true));
-        r.assertLogContains("My name on master is master using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+        r.assertLogContains("My name on master is master using labels " + builtInNodeLabel, r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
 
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -46,14 +46,15 @@
...
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "workflow-test");
 
         p.setDefinition(new CpsFlowDefinition(
-            "node('master') {\n" +
+            "node('" + builtInNodeLabel + "') {\n" +
             "  echo \"My name on master is ${env.NODE_NAME} using labels ${env.NODE_LABELS}\"\n" +
             "}\n",
             true));
-        r.assertLogContains("My name on master is master using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+        r.assertLogContains("My name on master is master using labels " + builtInNodeLabel, r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
```
daniel-beck, line 57: This is awkward; I missed `NODE_NAME` being defined in https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/cffc780c39172277ecde560eeba25a833c419740/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L812 so Freestyle and Pipeline are out of sync now.

daniel-beck: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/168

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -46,14 +46,15 @@
      */
     @Test public void isNodeNameAvailable() throws Exception {
         r.createSlave("node-test", "unix fast", null);
+        String builtInNodeLabel = r.jenkins.getSelfLabel().getExpression(); // compatibility with 2.307+
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "workflow-test");
 
         p.setDefinition(new CpsFlowDefinition(
-            "node('master') {\n" +
+            "node('" + builtInNodeLabel + "') {\n" +
```
jglick, line 53:
```suggestion
            "node('built-in || master') {\n" +
```

would also work right?

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
daniel-beck: Yes.

car-roll (approved):

PR #168, opened 2021-09-07
Change NODE_NAME for the built-in node based on its self label

daniel-beck:
This implements the same terminology cleanup as https://github.com/jenkinsci/jenkins/pull/5425 did in core for for `NODE_NAME` variable which is set here independently of core.

This is independent of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/167 but without that PR, this one will timeout on 2.307+ (unchanged from before).

This is not expected to result in a behavior difference on Jenkins 2.306 and earlier.

On Jenkins 2.307 and later, this PR will change the behavior and make it consistent with core's (freestyle jobs etc.). This can easily be tested by using the script console to provide a custom name and label for the built-in node:

```
Jenkins.nodeNameAndSelfLabelOverride = 'mast-in'
```

----

Changelog suggestion:

* The `NODE_NAME` environment variable value set for the built-in node inside `node` blocks now matches the behavior of other job types: On Jenkins 2.306 and earlier, the behavior is unchanged (`master`). On Jenkins 2.307 and later, the variable will be `built-in` or `master`, depending on whether the instance was upgraded from a previous release or newly installed, and whether the migration was applied. See https://www.jenkins.io/redirect/built-in-node-migration/

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,7 @@
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
+        <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
@@ -47,13 +47,14 @@ public class EnvWorkflowTest {
     @Test public void isNodeNameAvailable() throws Exception {
         r.createSlave("node-test", "unix fast", null);
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "workflow-test");
+        final String builtInNodeName = r.jenkins.getSelfLabel().getName();
 
         p.setDefinition(new CpsFlowDefinition(
             "node('master') {\n" +
             "  echo \"My name on master is ${env.NODE_NAME} using labels ${env.NODE_LABELS}\"\n" +
             "}\n",
             true));
-        r.assertLogContains("My name on master is master using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+        r.assertLogContains("My name on master is " + builtInNodeName + " using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
 
         p.setDefinition(new CpsFlowDefinition(
             "node('node-test') {\n" +
--- a/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
@@ -55,7 +55,7 @@ public class EnvWorkflowTest {
             "  echo \"My name on master is ${env.NODE_NAME} using labels ${env.NODE_LABELS}\"\n" +
             "}\n",
             true));
-        r.assertLogContains("My name on master is " + builtInNodeName + " using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+        r.assertLogContains("My name on master is " + builtInNodeName + " using labels " + builtInNodeLabel, r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
 
         p.setDefinition(new CpsFlowDefinition(
             "node('node-test') {\n" +
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -809,7 +809,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         env.put(COOKIE_VAR, cookie);
                         // Cf. CoreEnvironmentContributor:
                         if (exec.getOwner() instanceof MasterComputer) {
-                            env.put("NODE_NAME", "master");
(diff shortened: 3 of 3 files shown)


src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -47,13 +47,14 @@
...
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "workflow-test");
+        final String builtInNodeName = r.jenkins.getSelfLabel().getName();
 
         p.setDefinition(new CpsFlowDefinition(
             "node('master') {\n" +
             "  echo \"My name on master is ${env.NODE_NAME} using labels ${env.NODE_LABELS}\"\n" +
             "}\n",
             true));
-        r.assertLogContains("My name on master is master using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
+        r.assertLogContains("My name on master is " + builtInNodeName + " using labels master", r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
```
jglick, line 57: So was `Jenkins.getAssignedLabels` not updated?

jglick (commented):
Does https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/e53cdd6a350b93a40f0ad744857732f45a3a0da3/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L804-L805 still work?

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
daniel-beck: That's https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/167

jglick: Sorry, I am lost. Is there a reason why there are two PRs adapting to the same upstream change and editing the same line of code in different ways, meaning whoever merges them will be faced with a merge conflict on the second one?

daniel-beck:
> Does
> 
> https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/e53cdd6a350b93a40f0ad744857732f45a3a0da3/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L804-L805
> still work?

That's the "agent" branch, `computer.getName()` is weird for the built-in executor, always has been: https://github.com/jenkinsci/jenkins/blob/89d334145d2755f74f82aad07b5df4119d7fa6ce/core/src/main/java/jenkins/model/Jenkins.java#L5092-L5095

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
daniel-beck: #167 adapts the test to upstream changes to make PCT work and can basically be applied blindly. This PR changes behavior in quite significant ways. This makes them independently applicable (admittedly conflicting in an easy to resolve manner).

pom.xml
```diff
@@ -71,6 +71,7 @@
         <java.level>8</java.level>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
+        <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
```
daniel-beck, line 74: Not strictly correct but the best we can do to encourage admins to read the changelog and understand the impact of upgrading.

jglick:
> That's the "agent" branch

The code I linked to would run on the master node too.

daniel-beck:
> The code I linked to would run on the master node too.

You're technically correct (the best kind of correct), it's just that the return value of the getter would then be ignored, only line 814 uses it.

jglick:
> only line 814 uses it

No, it updates this field: https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/cffc780c39172277ecde560eeba25a833c419740/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L315-L316

daniel-beck:
Good point, I missed that 😞

Still, it's the same before and after for the built-in node: the empty string. Which is probably why the special behavior exists in the first place, users expect a non-empty `NODE_NAME`.

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
car-roll: it looks like this PR is also needed to make sure that the PCT does not run into the same timeout issues for 2.307+

car-roll (approved):

car-roll:
@jglick do you think your comments fully addressed?

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
jglick:
```suggestion
        r.assertLogContains("My name on master is " + builtInNodeName + " using labels " + builtInNodeLabel, r.assertBuildStatusSuccess(p.scheduleBuild2(0)));
```

?

jglick:
> do you think your comments fully addressed?

Other than https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/168#discussion_r707488266 yes.

timja (approved):
could we commit https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/168/files#r707488266 and 🚢  ?

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
daniel-beck: Looks like a bad merge between the PRs since both change different parts of this line. Checking locally...

daniel-beck:
@car-roll FYI I updated https://github.com/jenkinsci/workflow-durable-task-step-plugin/releases/tag/workflow-durable-task-step-2.40 in response to https://groups.google.com/g/jenkinsci-users/c/R7AUrb6nq5k/m/j_Omo1f3BQAJ; feel free to update/revert.

PR #170, opened 2021-09-16
Implement `Queue.Executable.getParentExecutable`

jglick:
Pairs with https://github.com/jenkinsci/jenkins/pull/5733; for now I am not bothering to increase the `jenkins.version`.

PR #177, opened 2021-11-05
Demonstration that implementing `getParentExecutable` is necessary

jglick:
Proves that #170 is necessary to retain compatibility with some rare use cases after https://github.com/jenkinsci/jenkins/pull/5880: for example, Pipeline code coverage without https://github.com/jenkinsci/code-coverage-api-plugin/pull/245 or https://github.com/jenkinsci/cobertura-plugin/pull/130.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -26,10 +26,13 @@ package org.jenkinsci.plugins.workflow.support.steps;
 
 import com.gargoylesoftware.htmlunit.Page;
 import com.google.common.base.Predicate;
+import com.google.common.collect.ImmutableSet;
+import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.Functions;
 import hudson.model.Computer;
 import hudson.model.Executor;
+import hudson.model.FreeStyleProject;
 import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Node;
@@ -37,6 +40,7 @@ import hudson.model.Queue;
 import hudson.model.Result;
 import hudson.model.Run;
 import hudson.model.Slave;
+import hudson.model.TaskListener;
 import hudson.model.User;
 import hudson.model.labels.LabelAtom;
 import hudson.model.queue.CauseOfBlockage;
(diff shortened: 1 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1219,6 +1234,92 @@ public void getOwnerTaskPermissions() throws Throwable {
...
+    public void accessPermittedOnlyFromCurrentBuild() throws Throwable {
+        sessions.then(r -> {
+            // Adapted from RunningBuildFilePathFilterTest
+            ExtensionList.lookupSingleton(AdminWhitelistRule.class).setMasterKillSwitch(false);
+            WorkflowJob main = r.createProject(WorkflowJob.class, "main");
+            DumbSlave s = r.createOnlineSlave();
+            main.setDefinition(new CpsFlowDefinition("node('" + s.getNodeName() + "') {writeBack()}", true));
+            // Normal case: writing to our own build directory
+            WriteBackStep.controllerFile = new File(main.getBuildDir(), "1/stuff.txt");
+            r.buildAndAssertSuccess(main);
```
jglick, line 1248: Fails without `getParentExecutable` override.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1219,6 +1234,92 @@ public void getOwnerTaskPermissions() throws Throwable {
...
+            // Attacks:
+            WriteBackStep.legal = false;
+            // Writing to someone else’s build directory (covered by RunningBuildFilePathFilter)
+            FreeStyleProject other = r.createFreeStyleProject("other");
+            r.buildAndAssertSuccess(other);
+            WriteBackStep.controllerFile = new File(other.getBuildByNumber(1).getRootDir(), "hack");
+            r.buildAndAssertSuccess(main);
```
jglick, line 1255: Fails against older Jenkins versions without `RunningBuildFilePathFilter`.

jglick:
This is more `tests` (or `dependencies`) than `enhancement`.

car-roll:
makes sense. I just wanted some kind of indicator that the jenkins version was being bumped, but i can also just add that in the changelog as well

PR #180, opened 2021-11-16
[JENKINS-49707] Agent missing after controller restart to fail resumption of `node` step, not kill whole build

jglick:
[JENKINS-49707](https://issues.jenkins.io/browse/JENKINS-49707)

Upstream of ~https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/195 and~ https://github.com/jenkinsci/kubernetes-plugin/pull/1211. ~Subsumes #231. Subsumes #249.~

Relevant prior work: #104, https://github.com/jenkinsci/kubernetes-plugin/pull/461, #47, #48, #115, #101, https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/86, #175, #176, #179.

Plan of record:
- [X] Introduce `ExecutorStepDynamicContext` and associated machinery as a replacement for `ExecutorPickle` and other pickles (see below).
- [x] Check PCT for any downstream regressions associated with this switch.
- [X] Introduce an extension point allowing the `retry` step to be given configurable behavior.
- [X] Create an implementation which matches generic “agent failure” conditions.
- [X] Offer a refined version in `kubernetes` which takes into account pod status.
- [X] Offer a retry option for Declarative `agent` types `any` and `label` (i.e., generic) and `kubernetes`
- [X] Decide what to do about resuming inside `node` but outside `sh`.
- [x] Revalidate behavior in a K8s disaster recovery scenario.
- [x] Port #184 to the new code, as https://github.com/jenkinsci/jenkins/pull/5934 can be reproduced again.
- [x] Some error saving `queue.xml`: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180#issuecomment-1126451601
- [x] 5m timeout restoring node should count from when build resumed; otherwise it takes 10m to resume a build with two parallel nodes (both missing) and retry: https://github.com/jenkinsci/workflow-api-plugin/pull/221/commits/b1778a924b2db390a4f4aed23f01d36493ff95f7
- [X] Extend #249 to `ExecutorStepDynamicContext`

The idea is that users updating plugins would see a somewhat altered behavior for resumption inside `node` blocks by default, which I hope would go mostly unnoticed; and could adjust pipelines to look something like
```groovy
podTemplate(…) {
  retry(count: 3, conditions: [kubernetesAgent(), nonresumable()]) {
    node(POD_LABEL) {
      checkout scm
      sh 'make world'
    }
  }
}
```
or
```groovy
pipeline {
  agent {
    kubernetes {
      yaml: '…'
      retries: 3
    }
  }
  stages {
    stage('main') {
      steps {
        sh 'make world'
      }
    }
  }
}
```

----

The key is to avoid relying on `Pickle`s for `ExecutorStepExecution` to resume properly. (More precisely, pickles using `TryRepeatedly`; `SecretPickle` and the like are harmless in this context.) Instead we take explicit control over how the step is resumed and what happens when. The problem with `Pickle` here is that it is too magical; we would like to be able to detect after resumption that an agent is gone and then throw an exception at the Groovy level, which is not possible if the build is aborted at the CPS VM level due to an inability to rehydrate.

----

I had planned to offer a system property to revert to `ExecutorPickle` in case of catastrophic problems (along with a couple sanity tests of that mode) but after reviewing the changes to `ExecutorStepExecution` this does not seem straightforward: the major simplifications and fixes to it rely on moving away from `ExecutorPickle`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -364,7 +364,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
                         } else {
                             LOGGER.fine(() -> "rediscovering that " + node + " has been removed and timeout has expired");
                             listener().getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
-                            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause());
                         }
                     }
                     removedNodeDiscovered = 0; // something else; reset
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -702,7 +702,11 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
                 // Defined in Java Platform and protected, so should not happen.
                 throw new ExceptionInInitializerError(x);
             }
-            printStreamDelegate.setAccessible(true);
+            try {
+                printStreamDelegate.setAccessible(true);
+            } catch (/* TODO Java 11+ InaccessibleObjectException */RuntimeException x) {
+                LOGGER.log(Level.WARNING, "On Java 17 error handling is degraded unless `--add-opens java.base/java.io=ALL-UNNAMED` is passed to the agent", x);
+            }
         }
 
         private static final long serialVersionUID = 1L;
@@ -720,7 +724,12 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             try {
                 if (ps.getClass() == PrintStream.class) {
                     // Try to extract the underlying stream, since swallowing exceptions is undesirable and PrintStream.checkError is useless.
-                    OutputStream os = (OutputStream) printStreamDelegate.get(ps);
+                    OutputStream os = ps;
+                    try {
+                        os = (OutputStream) printStreamDelegate.get(ps);
+                    } catch (IllegalAccessException x) {
+                        LOGGER.log(Level.FINE, "using PrintStream rather than underlying FilterOutputStream.out", x);
+                    }
                     if (os == null) { // like PrintStream.ensureOpen
                         throw new IOException("Stream closed");
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -25,6 +25,8 @@
 package org.jenkinsci.plugins.workflow.steps.durable_task;
 
 import com.google.common.collect.ImmutableSet;
+import edu.umd.cs.findbugs.annotations.CheckForNull;
+import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.AbortException;
 import hudson.EnvVars;
@@ -60,8 +62,6 @@ import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import javax.annotation.CheckForNull;
-import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import jenkins.tasks.filters.EnvVarsFilterableBuilder;
 import jenkins.util.Timer;
@@ -410,14 +410,14 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             }
         }
 
-        private synchronized @Nonnull TaskListener listener() {
+        private synchronized @NonNull TaskListener listener() {
             if (newlineSafeTaskListener == null) {
                 newlineSafeTaskListener = new NewlineSafeTaskListener(_listener());
             }
             return newlineSafeTaskListener;
         }
 
-        private @Nonnull TaskListener _listener() {
+        private @NonNull TaskListener _listener() {
             TaskListener l;
             StepContext context = getContext();
             try {
@@ -451,6 +451,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             }
 
             // Similar to DecoratedTaskListener:
+            @NonNull
             @Override public synchronized PrintStream getLogger() {
                 if (logger == null) {
                     LOGGER.fine("creating filtered stream");
@@ -461,7 +462,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
                             super.write(b);
                             nl = b == '\n';
                         }
-                        @Override public void write(byte[] b, int off, int len) throws IOException {
+                        @Override public void write(@NonNull byte[] b, int off, int len) throws IOException {
                             super.write(b, off, len);
                             if (len > 0) {
                                 nl = b[off + len - 1] == '\n';
@@ -486,7 +487,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
 
         }
 
-        private @Nonnull Launcher launcher() throws IOException, InterruptedException {
+        private @NonNull Launcher launcher() throws IOException, InterruptedException {
             StepContext context = getContext();
             Launcher l = context.get(Launcher.class);
             if (l == null) {
@@ -495,7 +496,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             return l;
         }
 
-        @Override public void stop(final Throwable cause) throws Exception {
+        @Override public void stop(@NonNull final Throwable cause) throws Exception {
             causeOfStoppage = cause;
             FilePath workspace = getWorkspace();
             if (workspace != null) {
@@ -714,7 +715,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             this.listener = listener;
         }
 
-        @Override public void output(InputStream stream) throws Exception {
+        @Override public void output(@NonNull InputStream stream) throws Exception {
             PrintStream ps = listener.getLogger();
             try {
                 if (ps.getClass() == PrintStream.class) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ComputerPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ComputerPickle.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.support.pickles;
 
+import edu.umd.cs.findbugs.annotations.NonNull;
 import org.jenkinsci.plugins.workflow.pickles.Pickle;
 import com.google.common.util.concurrent.ListenableFuture;
 import hudson.Extension;
@@ -64,7 +65,8 @@ public class ComputerPickle extends Pickle {
 
     @Extension
     public static final class Factory extends SingleTypedPickleFactory<Computer> {
-        @Override protected Pickle pickle(Computer computer) {
+        @NonNull
+        @Override protected Pickle pickle(@NonNull Computer computer) {
             return new ComputerPickle(computer);
         }
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -31,6 +31,7 @@ import hudson.Extension;
 import hudson.FilePath;
 import hudson.model.Computer;
 import hudson.model.Executor;
+import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.remoting.VirtualChannel;
 import hudson.slaves.OfflineCause;
@@ -46,6 +47,7 @@ import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.DynamicContext;
+import org.jenkinsci.plugins.workflow.support.DefaultStepContext;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
@@ -214,6 +216,27 @@ public final class ExecutorStepDynamicContext implements Serializable {
 
         @Override Computer get(ExecutorStepDynamicContext c) {
             return c.executor.getOwner();
+        }
+
+    }
+
+    /**
+     * Need not use {@link Translator} since we can serve a {@link Node} even when offline.
+     * Overrides default behavior in {@link DefaultStepContext} which would delegate to {@link ComputerTranslator}.
+     */
+    @Extension public static final class NodeTranslator extends DynamicContext.Typed<Node> {
+
+        @Override protected Class<Node> type() {
+            return Node.class;
+        }
+
+        @Override protected Node get(DelegatedContext context) throws IOException, InterruptedException {
+            ExecutorStepDynamicContext c = context.get(ExecutorStepDynamicContext.class);
+            if (c == null) {
+                return null;
+            }
+            Jenkins j = Jenkins.get();
+            return c.node.isEmpty() ? j : j.getNode(c.node);
         }
 
     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -24,29 +24,32 @@
 
 package org.jenkinsci.plugins.workflow.support.steps;
 
-import com.google.common.util.concurrent.Futures;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
 import hudson.FilePath;
+import hudson.Util;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Node;
 import hudson.model.Queue;
+import hudson.model.Result;
+import hudson.model.TaskListener;
 import hudson.remoting.VirtualChannel;
 import hudson.slaves.OfflineCause;
 import hudson.slaves.WorkspaceList;
 import java.io.IOException;
 import java.io.Serializable;
 import java.util.concurrent.CancellationException;
-import java.util.concurrent.ExecutionException;
-import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.DynamicContext;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
+import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.support.DefaultStepContext;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -68,12 +71,10 @@ public final class ExecutorStepDynamicContext implements Serializable {
     final @NonNull ExecutorStepExecution.PlaceholderTask task;
     final @NonNull String node;
     private final @NonNull String path;
-    /** Non-null sometime after {@link #resume} if all goes well. */
+    /** Non-null after {@link #resume} if all goes well. */
     private transient @Nullable Executor executor;
-    /** Non-null sometime after {@link #resume} if all goes well. */
+    /** Non-null after {@link #resume} if all goes well. */
     private transient @Nullable WorkspaceList.Lease lease;
-    /** Non-null after {@link #resume}. */
-    private transient @Nullable Future<Queue.Executable> future;
 
     ExecutorStepDynamicContext(ExecutorStepExecution.PlaceholderTask task, WorkspaceList.Lease lease, Executor executor) {
         this.task = task;
@@ -83,7 +84,7 @@ public final class ExecutorStepDynamicContext implements Serializable {
         this.lease = lease;
     }
 
-    void resume() {
+    void resume(StepContext context) throws Exception {
         if (executor != null) {
             throw new IllegalStateException("Already resumed");
         }
@@ -93,53 +94,49 @@ public final class ExecutorStepDynamicContext implements Serializable {
             throw new IllegalStateException("queue refused " + task);
         }
         LOGGER.fine(() -> "scheduled " + item + " for " + path + " on " + node);
-        future = item.getFuture().getStartCondition();
+        TaskListener listener = context.get(TaskListener.class);
+        if (!node.isEmpty()) { // unlikely to be any delay for built-in node anyway
+            listener.getLogger().println("Waiting for reconnection of " + node + " before proceeding with build");
+        }
+        Queue.Executable exec;
+        try {
+            exec = item.getFuture().getStartCondition().get(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
+        } catch (TimeoutException x) {
+            listener.getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
+            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+        } catch (CancellationException x) {
+            LOGGER.log(Level.FINE, "ceased to wait for " + node, x);
+            return;
+        }
+        executor = Executor.of(exec);
+        if (executor == null) {
+            // TODO this could happen as a race condition if the executable takes <1s to run; how could that be prevented?
+            // Or can we schedule a placeholder Task whose Executable does nothing but return Executor.currentExecutor and then end?
+            throw new IOException(exec + " was scheduled but no executor claimed it");
+        }
+        Computer computer = executor.getOwner();
+        VirtualChannel channel = computer.getChannel();
+        if (channel == null) {
+            throw new IOException(computer + " is offline");
+        }
+        FilePath fp = new FilePath(channel, path);
+        // Since there is no equivalent to Lock.tryLock for WorkspaceList (.record would work but throws AssertionError and swaps the holder):
+        WorkspaceList.Lease _lease = computer.getWorkspaceList().allocate(fp);
+        if (_lease.path.equals(fp)) {
+            lease = _lease;
+        } else { // @2 or other variant, not what we expected to be able to lock without contention
+            _lease.release();
+            throw new IOException("JENKINS-37121: something already locked " + fp);
+        }
+        LOGGER.fine(() -> "fully restored for " + path + " on " + node);
     }
 
     private static abstract class Translator<T> extends DynamicContext.Typed<T> {
 
         @Override protected T get(DelegatedContext context) throws IOException, InterruptedException {
             ExecutorStepDynamicContext c = context.get(ExecutorStepDynamicContext.class);
-            if (c == null) {
+            if (c == null || c.lease == null) {
                 return null;
-            }
-            if (c.executor == null || c.lease == null) {
-                if (c.future == null) {
-                    throw new IOException("Attempt to look up context from ExecutorStepExecution which has not been resumed");
-                }
-                if (!c.future.isDone()) {
-                    LOGGER.fine(() -> "queue item has not been scheduled for " + c.path + " on " + c.node + ", declining to provide any " + type().getSimpleName());
-                    return null;
-                }
-                LOGGER.fine(() -> "waiting for queue item to be scheduled for " + c.path + " on " + c.node);
-                Queue.Executable exec;
-                try {
-                    exec = c.future.get(1, TimeUnit.SECONDS);
-                } catch (ExecutionException | TimeoutException | CancellationException x) {
-                    c.future = Futures.immediateFailedFuture(x); // TODO Java 11+ CompletableFuture.failedFuture
-                    throw new IOException(x);
-                }
-                c.executor = Executor.of(exec);
-                if (c.executor == null) {
-                    // TODO this could happen as a race condition if the executable takes <1s to run; how could that be prevented?
-                    // Or can we schedule a placeholder Task whose Executable does nothing but return Executor.currentExecutor and then end?
-                    throw new IOException(exec + " was scheduled but no executor claimed it");
-                }
-                Computer computer = c.executor.getOwner();
-                VirtualChannel channel = computer.getChannel();
-                if (channel == null) {
-                    throw new IOException(computer + " is offline");
-                }
-                FilePath fp = new FilePath(channel, c.path);
-                // Since there is no equivalent to Lock.tryLock for WorkspaceList (.record would work but throws AssertionError and swaps the holder):
-                WorkspaceList.Lease lease = computer.getWorkspaceList().allocate(fp);
-                if (lease.path.equals(fp)) {
-                    c.lease = lease;
-                } else { // @2 or other variant, not what we expected to be able to lock without contention
-                    lease.release();
-                    throw new IOException("JENKINS-37121: something already locked " + fp);
-                }
-                LOGGER.fine(() -> "fully restored for " + c.path + " on " + c.node);
             }
             return get(c);
         }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -29,7 +29,6 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Util;
-import hudson.init.InitMilestone;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Node;
@@ -88,13 +87,6 @@ public final class ExecutorStepDynamicContext implements Serializable {
     void resume(StepContext context) throws Exception {
         if (executor != null) {
             throw new IllegalStateException("Already resumed");
-        }
-        while (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED || Jenkins.get().isQuietingDown()) {
-            if (Jenkins.get().isTerminating()) {
-                throw new IllegalStateException("Jenkins is now shutting down");
-            }
-            LOGGER.fine(() -> "waiting to schedule task for " + path + " on " + node + " until Jenkins completes startup and is not in quiet mode");
-            Thread.sleep(100);
         }
         Queue.Item item = Queue.getInstance().schedule2(task, 0).getItem();
         if (item == null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -89,8 +89,11 @@ public final class ExecutorStepDynamicContext implements Serializable {
         if (executor != null) {
             throw new IllegalStateException("Already resumed");
         }
-        while (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED) {
-            LOGGER.fine(() -> "waiting to schedule task for " + path + " on " + node + " until Jenkins completes startup");
+        while (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED || Jenkins.get().isQuietingDown()) {
+            if (Jenkins.get().isTerminating()) {
+                throw new IllegalStateException("Jenkins is now shutting down");
+            }
+            LOGGER.fine(() -> "waiting to schedule task for " + path + " on " + node + " until Jenkins completes startup and is not in quiet mode");
             Thread.sleep(100);
         }
         Queue.Item item = Queue.getInstance().schedule2(task, 0).getItem();
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.support.steps;
 
+import com.google.common.util.concurrent.Futures;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
@@ -45,7 +46,6 @@ import java.util.logging.Logger;
 import jenkins.model.Jenkins;
 import org.jenkinsci.plugins.workflow.FilePathUtils;
 import org.jenkinsci.plugins.workflow.steps.DynamicContext;
-import org.jenkinsci.plugins.workflow.support.concurrent.Futures;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
@@ -114,7 +114,7 @@ public final class ExecutorStepDynamicContext implements Serializable {
                 try {
                     exec = c.future.get(1, TimeUnit.SECONDS);
                 } catch (ExecutionException | TimeoutException | CancellationException x) {
-                    c.future = Futures.immediateFailedFuture(x);
+                    c.future = Futures.immediateFailedFuture(x); // TODO Java 11+ CompletableFuture.failedFuture
                     throw new IOException(x);
                 }
                 c.executor = Executor.of(exec);
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -108,10 +108,10 @@ public final class ExecutorStepDynamicContext implements Serializable {
             exec = item.getFuture().getStartCondition().get(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
         } catch (TimeoutException x) {
             listener.getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
-            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause());
         } catch (CancellationException x) {
             LOGGER.log(Level.FINE, "ceased to wait for " + node, x);
-            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.QueueTaskCancelled());
+            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.QueueTaskCancelled());
         }
         executor = Executor.of(exec);
         if (executor == null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -111,7 +111,7 @@ public final class ExecutorStepDynamicContext implements Serializable {
             throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
         } catch (CancellationException x) {
             LOGGER.log(Level.FINE, "ceased to wait for " + node, x);
-            return;
+            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.QueueTaskCancelled());
         }
         executor = Executor.of(exec);
         if (executor == null) {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -29,6 +29,7 @@ import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Util;
+import hudson.init.InitMilestone;
 import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.Node;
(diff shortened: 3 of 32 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ComputerPickle.java
```diff
@@ -34,8 +34,9 @@
 /**
  * Reference to {@link Computer}
  *
- * @author Kohsuke Kawaguchi
+ * @deprecated Normally now done via {@link ExecutorStepDynamicContext}.
  */
+@Deprecated
```
dwnusbaum, line 39: Can the factories for these pickles be deleted or replaced with an implementation that just throws an exception if it encounters an object of the specified type? Or do you want to keep them around in case some other step or wild Groovy code is relying on them?

jglick: I plan to retain them for compatibility. Weird Pipeline script is one possibility. They could also be used in other Pipeline steps; I know `PushdStep` binds `FilePathDynamicContext`, so that is not affected, but there may be others. Seems harmless enough to leave them here.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -775,18 +763,57 @@ private static void finish(@CheckForNull final String cookie) {
...
+                    }
+                    execution.state = null;
+                }
+            }
+
+            @Override public void onFailure(StepContext context, Throwable t) {
+                try {
+                    if (execution != null) {
+                        TaskListener listener = context.get(TaskListener.class);
+                        if (ExtensionList.lookup(ExecutorStepRetryEligibility.class).stream().anyMatch(e -> e.shouldRetry(t, execution.state.node, execution.step.getLabel(), listener))) {
```
dwnusbaum, line 806: Any thoughts on whether this behavior should be something users are able to disable at a per-step level (e.g. `node(..., allowRetry: false)`)? Or that they have to explicitly enable so that the current behavior does not change unless they want it to?

For example, maybe I have two `node` steps in a build, one of them runs tests and so I am happy to let it retry over and over, but the other performs a release and I would prefer to not retry that one even if some implementation of `ExecutorStepRetryEligibility.shouldRetry` returns `true`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -775,18 +763,57 @@ private static void finish(@CheckForNull final String cookie) {
...
+                            finished(context);
+                            execution.start();
+                            return;
+                        }
+                        listener.getLogger().println("No plugin requested a retry of a failed node block running on " + execution.state.node);
```
dwnusbaum, line 811: nit: Seems kind of unusual to me to put this message in the build log for users, since as far as I understand it is the current and expected behavior.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Yeah I have not really decided here. Making it opt-in would degrade the value quite a bit because you could not expect to just restore a whole cluster from backup (or whatever) and expect builds to resume without Pipeline modifications. Making it opt-out on a per-`node` basis could make sense.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Perhaps so.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: Yeah, maybe we could add a system property that works as a global opt-out in case someone runs into issues after updating.

jglick: Something like that. The possibilities both for opt-in and opt-out include

* a system property
* some `GlobalConfiguration` setting
* a parameter to the `node` step (also e.g. `KubernetesDeclarativeAgent`)

And not just _whether_ to retry, but if so, some maximum number of retries (so 0 to disable).

Safest would of course be to make all of this stuff be disabled by default (opt-in) initially. However in order for retry to work across restarts, I needed to suppress the various `Pickle`s in favor of `ExecutorStepDynamicContext`, which is a rather intrusive change (most of this PR); and conditionally using `ExecutorStepDynamicContext` but otherwise using `Pickle`s seemed like a nightmare to understand and test. I am fairly well convinced that moving away from `Pickle`s (at least those that need `TryRepeatedly`) and passing control over resumption to `StepExecution.onResume` implementations is the right thing to do in general—this goes back to a debate I had with @kohsuke in 2014—but it is pretty high risk. I do not expect to open this up for merge until January.

jglick:
A non-durable task can still fail here with a `MissingContextVariableException: Required context class hudson.FilePath is missing` it seems.

jglick:
[JENKINS-30383](https://issues.jenkins.io/browse/JENKINS-30383) also very desirable for retry of non-durable steps.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: But in practice, if retry is normally accepted but in some case is rejected, you want to know why. In fact it might be necessary to print the exception message or something. Needs refinement.

jglick:
Encountered a deadlock. Not sure if it is related to what @dwnusbaum was trying to fix in #188.

<details>
<summary>End of thread dump</summary>

```
Found one Java-level deadlock:
=============================
"Handling GET /jenkins/ from … : Jetty (winstone)-291 View/index.jelly":
  waiting to lock monitor 0x00007f1a6812d928 (object 0x00000000d9dbc808, a hudson.model.RunMap),
  which is held by "Executor #-1 for Built-In Node"
"Executor #-1 for Built-In Node":
  waiting for ownable synchronizer 0x00000000d2b1e2d8, (a java.util.concurrent.locks.ReentrantLock$NonfairSync),
  which is held by "SuggestReviewSoon (label='job-name-18-xjh7w-lp2sk-k05q7') / Computer.threadPoolForRemoting [#24]"
"SuggestReviewSoon (label='job-name-18-xjh7w-lp2sk-k05q7') / Computer.threadPoolForRemoting [#24]":
  waiting to lock monitor 0x00007f1a6812d928 (object 0x00000000d9dbc808, a hudson.model.RunMap),
  which is held by "Executor #-1 for Built-In Node"

Java stack information for the threads listed above:
===================================================
"Handling GET /jenkins/ from … : Jetty (winstone)-291 View/index.jelly":
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:376)
	- waiting to lock <0x00000000d9dbc808> (a hudson.model.RunMap)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.search(AbstractLazyLoadRunMap.java:351)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.newestBuild(AbstractLazyLoadRunMap.java:281)
	at jenkins.model.lazy.LazyBuildMixIn.getLastBuild(LazyBuildMixIn.java:242)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:241)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:104)
	at hudson.model.Job.getIconColor(Job.java:1180)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getIconColor(WorkflowJob.java:281)
	at …
"Executor #-1 for Built-In Node":
	at sun.misc.Unsafe.park(Native Method)
	- parking to wait for  <0x00000000d2b1e2d8> (a java.util.concurrent.locks.ReentrantLock$NonfairSync)
	at java.util.concurrent.locks.LockSupport.park(LockSupport.java:175)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.parkAndCheckInterrupt(AbstractQueuedSynchronizer.java:836)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquireQueued(AbstractQueuedSynchronizer.java:870)
	at java.util.concurrent.locks.AbstractQueuedSynchronizer.acquire(AbstractQueuedSynchronizer.java:1199)
	at java.util.concurrent.locks.ReentrantLock$NonfairSync.lock(ReentrantLock.java:209)
	at java.util.concurrent.locks.ReentrantLock.lock(ReentrantLock.java:285)
	at hudson.model.Queue.schedule2(Queue.java:567)
	at hudson.model.Queue.schedule2(Queue.java:693)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepDynamicContext.resume(ExecutorStepDynamicContext.java:88)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution.onResume(ExecutorStepExecution.java:194)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ResumeStepExecutionListener$1.onSuccess(FlowExecutionList.java:266)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ResumeStepExecutionListener$1.onSuccess(FlowExecutionList.java:243)
	at com.google.common.util.concurrent.Futures$6.run(Futures.java:975)
	at org.jenkinsci.plugins.workflow.flow.DirectExecutor.execute(DirectExecutor.java:33)
	at com.google.common.util.concurrent.ExecutionList$RunnableExecutorPair.execute(ExecutionList.java:149)
	at com.google.common.util.concurrent.ExecutionList.add(ExecutionList.java:105)
	at com.google.common.util.concurrent.AbstractFuture.addListener(AbstractFuture.java:155)
	at com.google.common.util.concurrent.Futures.addCallback(Futures.java:985)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionList$ResumeStepExecutionListener.onResumed(FlowExecutionList.java:243)
	at org.jenkinsci.plugins.workflow.flow.FlowExecutionListener.fireResumed(FlowExecutionListener.java:84)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.onLoad(WorkflowRun.java:567)
	- locked <0x00000000d5c0a9f8> (a java.lang.Object)
	at hudson.model.RunMap.retrieve(RunMap.java:226)
	at hudson.model.RunMap.retrieve(RunMap.java:58)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:506)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.load(AbstractLazyLoadRunMap.java:488)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:386)
	- locked <0x00000000d9dbc808> (a hudson.model.RunMap)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.search(AbstractLazyLoadRunMap.java:351)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.newestBuild(AbstractLazyLoadRunMap.java:281)
	at jenkins.model.lazy.LazyBuildMixIn.getLastBuild(LazyBuildMixIn.java:242)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:241)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:104)
	at hudson.model.Job.getEstimatedDurationCandidates(Job.java:1027)
	at hudson.model.Job.getEstimatedDuration(Job.java:1053)
	at hudson.model.Run.getEstimatedDuration(Run.java:2496)
	at hudson.model.Executor.run(Executor.java:408)
"SuggestReviewSoon (label='job-name-18-xjh7w-lp2sk-k05q7') / Computer.threadPoolForRemoting [#24]":
	at jenkins.model.lazy.AbstractLazyLoadRunMap.getByNumber(AbstractLazyLoadRunMap.java:376)
	- waiting to lock <0x00000000d9dbc808> (a hudson.model.RunMap)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.search(AbstractLazyLoadRunMap.java:351)
	at jenkins.model.lazy.AbstractLazyLoadRunMap.newestBuild(AbstractLazyLoadRunMap.java:281)
	at jenkins.model.lazy.LazyBuildMixIn.getLastBuild(LazyBuildMixIn.java:242)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:241)
	at org.jenkinsci.plugins.workflow.job.WorkflowJob.getLastBuild(WorkflowJob.java:104)
	at hudson.model.Job.getEstimatedDurationCandidates(Job.java:1027)
	at hudson.model.Job.getEstimatedDuration(Job.java:1053)
	at hudson.model.Run.getEstimatedDuration(Run.java:2496)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask.getEstimatedDuration(ExecutorStepExecution.java:683)
	at hudson.model.queue.MappingWorksheet.<init>(MappingWorksheet.java:327)
	at hudson.model.queue.MappingWorksheet.<init>(MappingWorksheet.java:312)
	at hudson.model.Queue.maintain(Queue.java:1645)
	at com.cloudbees.opscenter.client.cloud.CloudImpl$3.run(CloudImpl.java:402)
	at jenkins.util.ContextResettingExecutorService$1.run(ContextResettingExecutorService.java:28)
	at jenkins.security.ImpersonatingExecutorService$1.run(ImpersonatingExecutorService.java:68)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)

Found 1 deadlock.
```

</details>

jglick:
I guess that is just JENKINS-67351, despite the thread dumps being somewhat different, so I will try https://github.com/jenkinsci/workflow-api-plugin/pull/188 + https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/188.

dwnusbaum:
I think https://github.com/jenkinsci/workflow-api-plugin/pull/188 will fix that deadlock, since it will prevent the `"Executor #-1 for Built-In Node"` thread from trying to resume `ExecutorStepExecution` on the thread that holds the `RunMap` lock. Does look obviously related to #188.

jglick:
Observed an anomalous state where `KubernetesRetryEligibility` had activated and decided to retry a `node` block, but apparently never actually scheduled a new pod for that purpose (perhaps Jenkins shut down very soon afterward), and then with every subsequent restart a new `PlaceholderTask` would accumulate, all blocked waiting for the self-label of the long-since-deleted pod. Seems like https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/8beb608fd704a6702e3e045ca80f2c99d4ee6366/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L793-L802 never ran?

jglick:
Hitting https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/9c71521636a66ba6a345d247f36e8f82fb590fb7/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L191-L194 on a recently-started build, which could explain https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/9c71521636a66ba6a345d247f36e8f82fb590fb7/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L219

jglick:
b65ff1c seems to have a downside: AFAICT this is causing some non-`DurableTask` steps which require `FilePath` (`pwd` for example) to fail after a restart when not inside `ws` or if `ws` is modified to not use `WorkspaceListLeasePickle`. Observed in `ExecutorStepTest.acquireWorkspace`. Currently trying to see what breaks in `workflow-basic-steps`; first need to make `PushdStepTest.restarting` work with an agent.

jglick:
https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180#issuecomment-1006980635 is a broader problem. For example a reasonable-looking

<details>
<summary>Revision to test</summary>

```diff
diff --git src/test/java/org/jenkinsci/plugins/workflow/steps/PushdStepTest.java src/test/java/org/jenkinsci/plugins/workflow/steps/PushdStepTest.java
index c39e923..0c063ba 100644
--- src/test/java/org/jenkinsci/plugins/workflow/steps/PushdStepTest.java
+++ src/test/java/org/jenkinsci/plugins/workflow/steps/PushdStepTest.java
@@ -58,7 +58,7 @@ public class PushdStepTest {
         sessions.then(j -> {
             j.createSlave("remote", null, null);
             WorkflowJob p = j.createProject(WorkflowJob.class, "p");
-            p.setDefinition(new CpsFlowDefinition("node('remote') {dir('subdir') {semaphore 'restarting'; echo(/now the pwd=${pwd()}/)}}", true));
+            p.setDefinition(new CpsFlowDefinition("node('remote') {dir('subdir') {semaphore 'restarting'; sh 'echo PWD=$PWD'; echo(/Groovy pwd=${pwd()}/)}}", true));
             WorkflowRun b = p.scheduleBuild2(0).getStartCondition().get();
             SemaphoreStep.waitForStart("restarting/1", b);
         });
@@ -71,7 +71,9 @@ public class PushdStepTest {
             j.waitOnline(agent);
             FilePath ws = agent.getWorkspaceFor(p);
             assertNotNull(ws);
-            j.assertLogContains("now the pwd=" + ws.child("subdir"), j.assertBuildStatusSuccess(j.waitForCompletion(b)));
+            j.assertBuildStatusSuccess(j.waitForCompletion(b));
+            j.assertLogContains("PWD=" + ws.child("subdir"), b);
+            j.assertLogContains("Groovy pwd=" + ws.child("subdir"), b);
         });
     }
 
```
</details>

still fails

<details>
<summary>Stack trace in build log</summary>

```
java.io.IOException: Unable to create live FilePath for remote
	at org.jenkinsci.plugins.workflow.support.steps.FilePathDynamicContext.get(FilePathDynamicContext.java:66)
	at org.jenkinsci.plugins.workflow.support.steps.FilePathDynamicContext.get(FilePathDynamicContext.java:48)
	at org.jenkinsci.plugins.workflow.steps.DynamicContext$Typed.get(DynamicContext.java:95)
	at org.jenkinsci.plugins.workflow.cps.ContextVariableSet.get(ContextVariableSet.java:139)
	at org.jenkinsci.plugins.workflow.cps.CpsThread.getContextVariable(CpsThread.java:137)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.doGet(CpsStepContext.java:298)
	at org.jenkinsci.plugins.workflow.support.DefaultStepContext.get(DefaultStepContext.java:75)
	at org.jenkinsci.plugins.workflow.steps.StepDescriptor.checkContextAvailability(StepDescriptor.java:263)
	at org.jenkinsci.plugins.workflow.cps.DSL.invokeStep(DSL.java:299)
	at …
```
</details>

jglick:
After reconsideration, my new plan here:
* Delete `ExecutorStepRetryEligibility` and revert those changes to `ExecutorStepExecution`. Keep `ExecutorStepDynamicContext` and associated machinery.
* Check PCT for any downstream regressions associated with the switch from `ExecutorPickle` to `ExecutorStepDynamicContext`.
* Offer a system property to revert to `ExecutorPickle` in case of catastrophic problems, along with a couple sanity tests of that mode.
* Introduce a `Describable` extension point in `workflow-basic-steps` (or `workflow-api`?) allowing the `retry` step to be given configurable behavior.
* Convert `isGenerallyEligible` into an implementation which matches generic “agent failure” conditions.
* Offer a refined version in `kubernetes` which takes into account pod status.

Then users updating plugins would see a somewhat altered behavior for resumption inside `node` blocks by default, which I hope would go mostly unnoticed; and could adjust pipelines to look something like
```groovy
podTemplate(…) {
  retry(condition: kubernetesPodFailure()) {
    node(POD_LABEL) {
      sh 'make world'
    }
  }
}
```
Of course there could be an option for various Declarative `agent` types for that too.

jglick:
<details>
<summary>Error in log after restarting a build from checkpoint</summary>

```
…	WARNING	jenkins.security.ClassFilterImpl#notifyRejected: com.cloudbees.groovy.cps.impl.FunctionCallEnv in file:/tmp/jenkins/plugins/workflow-cps/WEB-INF/lib/groovy-cps-1.32.jar might be dangerous, so rejecting; see https://www.jenkins.io/redirect/class-filter/
…	WARNING	hudson.model.Queue#save: Failed to write out the queue file /var/jenkins_home/queue.xml
java.lang.UnsupportedOperationException: Refusing to marshal com.cloudbees.groovy.cps.impl.FunctionCallEnv for security reasons; see https://www.jenkins.io/redirect/class-filter/
	at hudson.util.XStream2$BlacklistedTypesConverter.marshal(XStream2.java:582)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize com.cloudbees.groovy.cps.Continuable#e for class org.jenkinsci.plugins.workflow.cps.SandboxContinuable
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.cps.CpsThread#program for class org.jenkinsci.plugins.workflow.cps.CpsThread
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:44)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:87)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeBareItem(AbstractCollectionConverter.java:94)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeItem(AbstractCollectionConverter.java:66)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeCompleteItem(AbstractCollectionConverter.java:81)
	at com.thoughtworks.xstream.converters.collections.MapConverter.marshal(MapConverter.java:80)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.cps.CpsThreadGroup#threads for class org.jenkinsci.plugins.workflow.cps.CpsThreadGroup
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.cps.CpsThread#group for class org.jenkinsci.plugins.workflow.cps.CpsThread
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.cps.CpsBodyExecution#thread for class org.jenkinsci.plugins.workflow.cps.CpsBodyExecution
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution#body for class org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask#execution for class org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize hudson.model.Queue$Item#task for class hudson.model.Queue$BuildableItem
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:44)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:87)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeBareItem(AbstractCollectionConverter.java:94)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeItem(AbstractCollectionConverter.java:66)
	at com.thoughtworks.xstream.converters.collections.AbstractCollectionConverter.writeCompleteItem(AbstractCollectionConverter.java:81)
	at com.thoughtworks.xstream.converters.collections.CollectionConverter.marshal(CollectionConverter.java:75)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller$1.convertAnother(AbstractReferenceMarshaller.java:83)
	at hudson.util.RobustReflectionConverter.marshallField(RobustReflectionConverter.java:283)
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:270)
Caused: java.lang.RuntimeException: Failed to serialize hudson.model.Queue$State#items for class hudson.model.Queue$State
	at hudson.util.RobustReflectionConverter$2.writeField(RobustReflectionConverter.java:274)
	at hudson.util.RobustReflectionConverter$2.visit(RobustReflectionConverter.java:241)
	at com.thoughtworks.xstream.converters.reflection.PureJavaReflectionProvider.visitSerializableFields(PureJavaReflectionProvider.java:174)
	at hudson.util.RobustReflectionConverter.doMarshal(RobustReflectionConverter.java:226)
	at hudson.util.RobustReflectionConverter.marshal(RobustReflectionConverter.java:163)
	at com.thoughtworks.xstream.core.AbstractReferenceMarshaller.convert(AbstractReferenceMarshaller.java:68)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:59)
	at com.thoughtworks.xstream.core.TreeMarshaller.convertAnother(TreeMarshaller.java:44)
	at com.thoughtworks.xstream.core.TreeMarshaller.start(TreeMarshaller.java:83)
	at com.thoughtworks.xstream.core.AbstractTreeMarshallingStrategy.marshal(AbstractTreeMarshallingStrategy.java:37)
	at com.thoughtworks.xstream.XStream.marshal(XStream.java:1266)
	at com.thoughtworks.xstream.XStream.marshal(XStream.java:1255)
	at com.thoughtworks.xstream.XStream.toXML(XStream.java:1228)
	at hudson.XmlFile.write(XmlFile.java:213)
Caused: java.io.IOException
	at hudson.XmlFile.write(XmlFile.java:220)
	at hudson.model.Queue.save(Queue.java:470)
	at hudson.model.Queue$Saver.run(Queue.java:3121)
	at jenkins.security.ImpersonatingScheduledExecutorService$1.run(ImpersonatingScheduledExecutorService.java:67)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

</details>

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -274,22 +263,25 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
...
-                        listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        body.cancel(new RemovedNodeCause());
+                        task.withExecution(execution -> {
+                            BodyExecution body = execution.body;
+                            if (body == null) {
+                                listener.getLogger().println("Agent " + node.getNodeName() + " was deleted, but do not have a node body to cancel");
+                                return;
+                            }
+                            listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
+                            body.cancel(new RemovedNodeCause()); // TODO pass actualInterruption=false https://github.com/jenkinsci/workflow-cps-plugin/blob/f642ed4fad556e100394ebb1f69f0ac8e5336248/src/main/java/org/jenkinsci/plugins/workflow/cps/CpsBodyExecution.java#L251
```
jglick, line 280: https://github.com/jenkins-infra/helpdesk/issues/3031#issuecomment-1175343746

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -0,0 +1,246 @@
...
+        this.path = lease.path.getRemote();
+        this.executor = executor;
+        this.lease = lease;
+    }
+
+    void resume(StepContext context) throws Exception {
+        if (executor != null) {
+            throw new IllegalStateException("Already resumed");
+        }
+        while (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED) {
```
jtnord, line 92: when is this called? if Jenkins fails to start will this leave a non deamon thread running that causes the Jenkins termination to fail?

Should this have some form of an exit path for `Jenkins.get().isTerminating()`?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -0,0 +1,246 @@
...
+            throw new IllegalStateException("queue refused " + task);
+        }
+        LOGGER.fine(() -> "scheduled " + item + " for " + path + " on " + node);
+        TaskListener listener = context.get(TaskListener.class);
+        if (!node.isEmpty()) { // unlikely to be any delay for built-in node anyway
+            listener.getLogger().println("Waiting for reconnection of " + node + " before proceeding with build");
+        }
+        Queue.Executable exec;
+        try {
+            exec = item.getFuture().getStartCondition().get(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
```
jtnord, line 108: What if Jenkins is started in quiet down mode?

Does this mean that the job that was running will always fail?

jtnord (commented):
Can't say I am an expert in this code - there are 2 questions I have inline.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
jglick: > a non deamon thread running that causes the Jenkins termination to fail

Not sure offhand if that is actually true, but I guess it should not hurt to check for the terminating flag. Will try it.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -356,12 +358,13 @@ static final class Execution extends AbstractStepExecutionImpl implements Runnab
...
+                        } else if (System.nanoTime() - removedNodeDiscovered < TimeUnit.MILLISECONDS.toNanos(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS)) {
                             LOGGER.fine(() -> "rediscovering that " + node + " has been removed");
                             return null;
                         } else {
-                            LOGGER.fine(() -> node + " has been removed for a while, assuming it is not coming back");
-                            throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+                            LOGGER.fine(() -> "rediscovering that " + node + " has been removed and timeout has expired");
+                            listener().getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
+                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause());
```
jglick, line 367: Note that https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/97365d4727f1647c94dd11e4081085305239572f/src/main/java/org/jenkinsci/plugins/workflow/support/steps/AgentErrorCondition.java#L60-L62 bypasses the `actualInterruption` check in https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/d57e3ca46d24bb91b8410467284176b0f319c64e/src/main/java/org/jenkinsci/plugins/workflow/steps/RetryStepExecution.java#L93 so the distinction would only matter for other catching steps like https://github.com/jenkinsci/workflow-basic-steps-plugin/blob/d57e3ca46d24bb91b8410467284176b0f319c64e/src/main/java/org/jenkinsci/plugins/workflow/steps/CatchErrorStep.java#L229

src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
```diff
@@ -45,7 +45,7 @@
  * Allows a step body to save a representation of a workspace
  * without forcing a particular {@link FilePath#getChannel} to be used the whole time.
  */
-@Extension public final class FilePathDynamicContext extends DynamicContext.Typed<FilePath> {
+@Extension(ordinal = 100) public final class FilePathDynamicContext extends DynamicContext.Typed<FilePath> {
```
dwnusbaum, line 48: Is this needed for `PushdStep` to keep working, or something else?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -182,35 +193,13 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
-                        }
-                    }
-                }
-            }
-            TaskListener listener = getContext().get(TaskListener.class);
-            if (step == null) { // compatibility: used to be transient
-                listener.getLogger().println("Queue item for node block in " + run.getFullDisplayName() + " is missing (perhaps JENKINS-34281), but cannot reschedule");
+            if (state == null) {
+                Run<?, ?> run = getContext().get(Run.class);
+                LOGGER.fine(() -> "No ExecutorStepDynamicContext found for node block in " + run + "; perhaps loading from a historical build record, hoping for the best");
```
dwnusbaum, line 199: What is the expected behavior in this case? The various serialized pickles are rehydrated and the behavior is the same as it was before? Do we have any `@LocalData` tests that cover this case?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -376,6 +357,20 @@ private Object readResolve() {
             return this;
         }
 
+        /**
+         * We cannot keep {@link ExecutorStepExecution} as a serial field of {@link PlaceholderTask}
+         * since it could not be serialized via XStream in {@link Queue}.
+         * Instead we keep only {@link #context} and look up the execution as needed.
+         */
+        private void withExecution(Consumer<ExecutorStepExecution> executionCallback) {
+            StepExecution.applyAll(ExecutorStepExecution.class, execution -> {
```
dwnusbaum, line 366: This iterates over all running steps in all running Pipelines, right? Why not look up the `FlowExecution` from `context` instead so that we can call `FlowExecution.getCurrentExecutions` directly to only consider steps for this build? Maybe just because this is expected to run infrequently and so performance is not a concern?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -794,18 +789,42 @@ private static void finish(@CheckForNull final String cookie) {
         private static final class Callback extends BodyExecutionCallback.TailCall {
```
dwnusbaum, line 789: Do we need to add an explicit `serialVersionUID` to this class to avoid compatibility issues since a field has been added?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -794,18 +789,42 @@ private static void finish(@CheckForNull final String cookie) {
...
             }
 
             @Override protected void finished(StepContext context) throws Exception {
                 LOGGER.log(FINE, "finished {0}", cookie);
-                lease.release();
-                lease = null;
-                finish(cookie);
+                try {
+                    if (execution != null) {
+                        WorkspaceList.Lease _lease = ExtensionList.lookupSingleton(ExecutorStepDynamicContext.WorkspaceListLeaseTranslator.class).get(execution.state);
```
dwnusbaum, line 805: Why can't this be `WorkspaceList.Lease _lease = context.get(WorkspaceList.Lease);`? Because `ExecutorStepDynamicContext` has already been dropped from the context once we enter the callback?

dwnusbaum (commented):
I took a brief look at the PR and added some questions, but in general it looks fine. I am planning on reviewing it more thoroughly tomorrow.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/FilePathDynamicContext.java
jglick: Yes exactly that.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: The only difference in behavior should be for the cases formerly called out below with the JENKINS-34281 warning. In normal cases, this method would have `return`ed earlier. So I decided not to bother keeping the complicated code which should only apply after the one-time upgrade and even then only if something weird was broken.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Cannot recall whether I tried that. Can check if it works.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: That would probably make sense.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Cannot recall. I think this is the wrong context?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: Ok, seems fine then. I think it would be nice though to have a happy-path `@LocalData` test to verify behavior across the upgrade, even if we just keep it for a few weeks and then delete it.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -0,0 +1,249 @@
...
+        this.path = lease.path.getRemote();
+        this.executor = executor;
+        this.lease = lease;
+    }
+
+    void resume(StepContext context) throws Exception {
+        if (executor != null) {
+            throw new IllegalStateException("Already resumed");
+        }
+        while (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED || Jenkins.get().isQuietingDown()) {
```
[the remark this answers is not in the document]
jglick: Yes, I think that would make sense. Could be done before merging this PR I think.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Checked, and `context.get(WorkspaceList.Lease.class)` is indeed null here. `context` is a `CpsBodySubContext`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: https://github.com/jenkinsci/workflow-cps-plugin/blob/f82abbe3e8a5c749f05b2ad77d0707a0a6d0db0a/src/test/java/org/jenkinsci/plugins/workflow/cps/SerialFormTest.java#L76-L98 covers this code path.

dwnusbaum (approved):

dwnusbaum (approved):

basil:
Breaks `io.jenkins.docker.pipeline.DockerNodeStepTest#withinNode`, `io.jenkins.docker.pipeline.DockerNodeStepTest#defaults`, and `io.jenkins.docker.pipeline.DockerNodeStepTest#nodeWithinDockerNodeWithinNode` as discovered in https://github.com/jenkinsci/docker-plugin/pull/943.

PR #184, opened 2021-11-19
`ExecutorPickle` should not reschedule anything until after `Queue.load`

jglick:
See https://github.com/jenkinsci/jenkins/pull/5934.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -88,7 +88,7 @@ public class ExecutorPickle extends Pickle {
     @Override public ListenableFuture<Executor> rehydrate(final FlowExecutionOwner owner) {
         return new TryRepeatedly<Executor>(1, 0) {
             long itemID;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -91,6 +92,10 @@ private ExecutorPickle(Executor executor) {
 
             @Override
             protected Executor tryResolve() throws Exception {
+                if (Jenkins.get().getInitLevel() != InitMilestone.COMPLETED) {
+                    LOGGER.fine(() -> "not going to schedule " + task + " yet because Jenkins has not yet completed startup");
+                    return null;
```
dwnusbaum, line 97: Will any retries due to this condition count against the 5-minute or so timer before the build gets aborted automatically? Or does that only apply once the build is in the queue?

jglick: Currently it will. You are right, it would be better to start the timer when we schedule the item.

dwnusbaum (approved):

car-roll (approved):

PR #185, opened 2021-12-01
Recover gracefully when a `PlaceholderTask` is in the queue but the associated build is complete

dwnusbaum:
While testing Jenkins in various backup and restore scenarios, we ran into a case where the Jenkins queue contained a `ExecutorStepExecution.PlaceholderTask` whose associated build was already complete, causing the task to sit in the queue forever. We are not sure how this happened or how to reproduce it, but this PR adds a test that reproduces the same symptoms that may be useful for future investigation.

I did some experimentation to see if the issue could be fixed, and something like this does fix it but does not seem like the best approach:

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
index c81bf0c..64ca128 100644
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -360,6 +360,10 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         }
 
         private Object readResolve() {
+            Run<?, ?> run = runForDisplay();
+            if (run != null && !run.isLogUpdated()) {
+                return null;
+            }
             if (cookie != null) {
                 synchronized (runningTasks) {
                     runningTasks.put(cookie, new RunningTask());
```

A patch like this would cause Pipeline builds to be loaded and resumed from inside of `Queue.load` while the queue lock is held during Jenkins startup, which seems concerning. Accessing any context variables via `StepContext.get` would have the same problem because `CpsStepContext.doGet` transitively calls `getThreadGroupSynchronously` which has to load the `WorkflowRun` and resume the Pipeline. Perhaps we could expose `CpsStepContext.isComplete` as a new method on `StepContext` and check that instead, but I'm not sure. `Queue.load` [does explicitly handle the case where a task is null](https://github.com/jenkinsci/jenkins/blob/da7935d6cca1e48adbdea9ecae8c9a150e8ce5e8/core/src/main/java/hudson/model/Queue.java#L411), although maybe there is a better way to remove the task from the queue.

Given @jglick is working on stuff related to resumption of the `node` step in #180, it seems best not to try to make a speculative fix right now.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -71,6 +71,7 @@
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
+        <jenkins-test-harness.version>1664.ve9ed23f5e0f2</jenkins-test-harness.version> <!-- TODO: https://github.com/jenkinsci/jenkins-test-harness/pull/353 -->
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -71,7 +71,7 @@
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
-        <jenkins-test-harness.version>1664.ve9ed23f5e0f2</jenkins-test-harness.version> <!-- TODO: https://github.com/jenkinsci/jenkins-test-harness/pull/353 -->
+        <jenkins-test-harness.version>1666.vd1360abbfe9e</jenkins-test-harness.version>
     </properties>
     <dependencyManagement>
         <dependencies>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -426,9 +426,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             if (!stopping && run != null && !run.isLogUpdated()) {
                 stopping = true;
                 LOGGER.warning(() -> "Refusing to build " + this + " and cancelling it because associated build is complete");
-                Timer.get().execute(() -> {
-                    Queue.getInstance().cancel(this);
-                });
+                Timer.get().execute(() -> Queue.getInstance().cancel(this));
             }
             if (stopping) {
                 return new CauseOfBlockage() {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -422,6 +422,22 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         }
 
         @Override public CauseOfBlockage getCauseOfBlockage() {
+            Run<?, ?> run = runForDisplay();
+            if (!stopping && run != null && !run.isLogUpdated()) {
+                stopping = true;
+                LOGGER.warning(() -> "Refusing to build " + this + " and cancelling it because associated build is complete");
+                Timer.get().execute(() -> {
+                    Queue.getInstance().cancel(this);
+                });
+            }
+            if (stopping) {
+                return new CauseOfBlockage() {
+                    @Override
+                    public String getShortDescription() {
+                        return "Stopping " + getDisplayName();
+                    }
+                };
+            }
             return null;
         }
 
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -35,6 +35,7 @@ import hudson.model.Executor;
 import hudson.model.FreeStyleProject;
 import hudson.model.Item;
 import hudson.model.Job;
+import hudson.model.Label;
 import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.Result;
@@ -79,7 +80,9 @@ import javax.annotation.Nullable;
 
 import hudson.util.VersionNumber;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.StandardCopyOption;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicReference;
 import jenkins.model.Jenkins;
 import jenkins.security.MasterToSlaveCallable;
 import jenkins.security.QueueItemAuthenticator;
@@ -126,6 +129,7 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import org.junit.Assume;
 import org.junit.ClassRule;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -1264,6 +1268,44 @@ public class ExecutorStepTest {
             r.buildAndAssertSuccess(main);
         });
     }
+
+    @Ignore
+    @Test public void placeholderTaskInQueueButAssociatedBuildComplete() throws Throwable {
+        AtomicReference<File> rootDir = new AtomicReference<>();
+        Path tempQueueFile = Files.createTempFile("queue", ".xml");
+        sessions.then(r -> {
+            rootDir.set(r.jenkins.getRootDir());
+            System.out.println(rootDir);
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("node('custom-label') { }", true));
+            WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+            // Get into a state where a PlaceholderTask is in the queue.
+            while (true) {
+                Queue.Item[] items = Queue.getInstance().getItems();
+                if (items.length == 1 && items[0].task instanceof ExecutorStepExecution.PlaceholderTask) {
+                    break;
+                }
+                Thread.sleep(500L);
+            }
+            // Copy queue.xml to a temp file while the PlaceholderTask is in the queue.
+            r.jenkins.getQueue().save();
+            Files.copy(rootDir.get().toPath().resolve("queue.xml"), tempQueueFile, StandardCopyOption.REPLACE_EXISTING);
+            // Create a node with the correct label and let the build complete.
+            DumbSlave node = r.createOnlineSlave(Label.get("custom-label"));
+            r.assertBuildStatusSuccess(r.waitForCompletion(b));
+        });
+        // Copy the temp queue.xml over the real one. The associated build has already completed, so the queue now
+        // has a bogus PlaceholderTask.
+        Files.copy(tempQueueFile, rootDir.get().toPath().resolve("queue.xml"), StandardCopyOption.REPLACE_EXISTING);
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            assertFalse(b.isLogUpdated());
+            r.assertBuildStatusSuccess(b);
+            assertThat(Queue.getInstance().getItems(), emptyArray()); // This assertion fails.
+        });
+    }
+
     public static final class WriteBackStep extends Step {
         static File controllerFile;
         static boolean legal = true;
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -1269,7 +1269,7 @@ public class ExecutorStepTest {
         });
     }
 
-    @Ignore
+    @Ignore("TODO safe fix still TBD")
     @Test public void placeholderTaskInQueueButAssociatedBuildComplete() throws Throwable {
         AtomicReference<File> rootDir = new AtomicReference<>();
         Path tempQueueFile = Files.createTempFile("queue", ".xml");
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -128,7 +128,6 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 import org.junit.Assume;
 import org.junit.ClassRule;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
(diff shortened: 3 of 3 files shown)


jglick:
I need to think about the interaction with #180.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1264,6 +1268,44 @@ public void accessPermittedOnlyFromCurrentBuild() throws Throwable {
             r.buildAndAssertSuccess(main);
         });
     }
+
+    @Ignore
```
jglick, line 1272:
```suggestion
    @Ignore("TODO safe fix still TBD")
```

jglick (approved):

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1264,6 +1268,44 @@ public void accessPermittedOnlyFromCurrentBuild() throws Throwable {
             r.buildAndAssertSuccess(main);
         });
     }
+
+    @Ignore("TODO safe fix still TBD")
+    @Test public void placeholderTaskInQueueButAssociatedBuildComplete() throws Throwable {
+        AtomicReference<File> rootDir = new AtomicReference<>();
```
dwnusbaum, line 1274: Unrelated, but this could be simplified if we exposed `JenkinsSessionRule.home` via a getter. Any thoughts on that? Back when we only had `RestartableJenkinsRule` this kind of test would've been more awkward to write, but now that we can do things to `JENKINS_HOME` between restarts with `JenkinsSessionsRule` it's pretty straightforward.

jglick: Funny you should mention that: https://github.com/jenkinsci/jenkins-test-harness/pull/352#discussion_r760545577

dwnusbaum: See 3abef26 (also switched to `TemporaryFolder` and removed a `System.out.println` that was not meant to be committed).

pom.xml
```diff
@@ -71,6 +71,7 @@
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
+        <jenkins-test-harness.version>1664.ve9ed23f5e0f2</jenkins-test-harness.version> <!-- TODO: https://github.com/jenkinsci/jenkins-test-harness/pull/353 -->
```
jglick, line 74: update to release version

dwnusbaum:
Moving back to draft to wait for a release of https://github.com/jenkinsci/jenkins-test-harness/pull/353 and to investigate https://github.com/jenkinsci/workflow-cps-plugin/pull/490#issuecomment-984976017.

jglick (approved):
A `QueueDecisionHandler` might fix the real bug, but not the problem path tested here. You could add a `QueueTaskDispatcher` whose `canRun` cancels the item if it matches this condition, which avoids the problems with `readResolve` but on the other hand adds overhead to a critical code path for what I guess is a rare case. Maybe you could implement a fix in `PlaceholderTask.getAssignedLabel` or `getAffinityKey` or some similar method `Queue.maintain` is guaranteed to call, cancelling itself (perhaps asynchronously via `Timer` to avoid weird reëntrancy issues)?

dwnusbaum:
FWIW I looked into exposing `CpsStepContext.isCompleted` as an API on `StepContext` to be able to check if the step completed without having to load the build in `PlaceholderTask.readResolve`, but then I realized that kind of fix would be useless for my test case since the serialized `CpsStepContext` would be outdated and so `isCompleted` would return `false`.

> Maybe you could implement a fix in PlaceholderTask.getAssignedLabel or getAffinityKey or some similar method Queue.maintain is guaranteed to call, cancelling itself (perhaps asynchronously via Timer to avoid weird reëntrancy issues)?

Yeah I was thinking about something like this. I will spend a bit of time looking into it.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1264,6 +1267,41 @@ public void accessPermittedOnlyFromCurrentBuild() throws Throwable {
...
+        });
+        // Copy the temp queue.xml over the real one. The associated build has already completed, so the queue now
+        // has a bogus PlaceholderTask.
+        Files.copy(tempQueueFile, sessions.getHome().toPath().resolve("queue.xml"), StandardCopyOption.REPLACE_EXISTING);
+        sessions.then(r -> {
+            WorkflowJob p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+            WorkflowRun b = p.getBuildByNumber(1);
+            assertFalse(b.isLogUpdated());
+            r.assertBuildStatusSuccess(b);
+            assertThat(Queue.getInstance().getItems(), emptyArray()); // This assertion fails.
```
dwnusbaum, line 1301: Turns out the test as-is is flaky right now. If you add a 5 second sleep before this line so that the agent has time to come online, then the placeholder task attempts to run, fails, and the assertion succeeds.

Specifically [this call to `StepContext.get`](https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/f832bc1ac7457afe18fb6d62b199c7f6c29335cf/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L802) fails with this exception because it expects the execution to still be in progress when it has already completed:

```
java.lang.IllegalStateException: completed or broken execution
	at org.jenkinsci.plugins.workflow.cps.CpsFlowExecution.onLoad(CpsFlowExecution.java:764)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun.getExecution(WorkflowRun.java:691)
	at org.jenkinsci.plugins.workflow.job.WorkflowRun$Owner.get(WorkflowRun.java:963)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.getExecution(CpsStepContext.java:217)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.getThreadGroupSynchronously(CpsStepContext.java:242)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.access$000(CpsStepContext.java:97)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext$1.call(CpsStepContext.java:263)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext$1.call(CpsStepContext.java:261)
	at jenkins.util.ContextResettingExecutorService$2.call(ContextResettingExecutorService.java:46)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
Caused: java.util.concurrent.ExecutionException
	at com.google.common.util.concurrent.AbstractFuture$Sync.getValue(AbstractFuture.java:289)
	at com.google.common.util.concurrent.AbstractFuture$Sync.get(AbstractFuture.java:276)
	at com.google.common.util.concurrent.AbstractFuture.get(AbstractFuture.java:111)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.getThreadGroupSynchronously(CpsStepContext.java:247)
Caused: java.io.IOException
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.getThreadGroupSynchronously(CpsStepContext.java:249)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.getThreadSynchronously(CpsStepContext.java:236)
	at org.jenkinsci.plugins.workflow.cps.CpsStepContext.doGet(CpsStepContext.java:293)
	at org.jenkinsci.plugins.workflow.support.DefaultStepContext.get(DefaultStepContext.java:75)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask$PlaceholderExecutable.run(ExecutorStepExecution.java:806)
	at hudson.model.ResourceController.execute(ResourceController.java:99)
```

Somewhat interestingly, this does not trigger the "already completed" log messages from `CpsStepContext.completed` because the copy of `CpsStepContext` in question was serialized before the step completed and so it thinks it is completing for the first time. This means that no warnings are logged at all when this happens, we just save the above exception into `CpsStepContext.outcome` and then `CpsStepContext.scheduleNextRun` fails silently in the `onFailure` arm of the future passed to `runInCpsVmThread`.

I can de-flake the test by removing the node before the restart, which I think is still somewhat analogous to what might happen with a non-static agent after a restart, but I'm not sure.

jglick: > I can de-flake the test by removing the node before the restart, which I think is still somewhat analogous to what might happen with a non-static agent after a restart

:+1:

dwnusbaum: De-flaked in 00cdd69.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -422,6 +422,22 @@ public String getCookie() {
         }
 
         @Override public CauseOfBlockage getCauseOfBlockage() {
+            Run<?, ?> run = runForDisplay();
```
dwnusbaum, line 425: I think this is safe. If we get here for a step that is just starting or resuming, then the run is already loaded and so this should complete quickly. The only time this should be slow is if this is after a Jenkins restart and the build has already completed so we end up here without the build having been loaded via some other route and we trigger the cancellation path.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -422,6 +422,22 @@ public String getCookie() {
...
+                    Queue.getInstance().cancel(this);
+                });
+            }
+            if (stopping) {
+                return new CauseOfBlockage() {
+                    @Override
+                    public String getShortDescription() {
+                        return "Stopping " + getDisplayName();
+                    }
+                };
```
dwnusbaum, line 439: Not sure if an anonymous class or hard-coded text is ok here. My thought was that this cause should not usually be around long enough for anyone to see it, but I guess we should set up localization just in case.

dwnusbaum:
@jglick let me know what you think of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185/commits/9d8eeca06ba0eab24dea6ff8990dcc7ef0770c3a. If you think there are some problems with that approach then I'll just revert that commit and leave the test ignored for now.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -422,6 +422,22 @@ public String getCookie() {
         }
 
         @Override public CauseOfBlockage getCauseOfBlockage() {
+            Run<?, ?> run = runForDisplay();
+            if (!stopping && run != null && !run.isLogUpdated()) {
+                stopping = true;
+                LOGGER.warning(() -> "Refusing to build " + this + " and cancelling it because associated build is complete");
+                Timer.get().execute(() -> {
+                    Queue.getInstance().cancel(this);
+                });
```
jglick, line 431:
```suggestion
                Timer.get().execute(() -> Queue.getInstance().cancel(this));
```

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: An anonymous class should be fine here. As you say, it ought not appear in the GUI for more than a moment if at all.

jglick (approved):
Looks reasonable to me.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: Ah yeah, at one point I was logging something in the block and I forgot to simplify it.

car-roll (approved):

dduportal:
The issue described (and fixed) by this PR happens quite often on restarts of ci.jenkins.io (but not on other instances owned by the infra team).
Would you want us to try this change on ci.jenkins in  advance? I'm asking because I'm not sure to really get all the impacts that it could have, but since ci.jenkins.io is basically a giant infinite retry loop for its jobs, that should not cause harm while allowing the "rare" bug to be caught and tried. Let us know?

jglick:
@dduportal sure, if you want to install an incremental build (assuming all dependencies are met) and be prepared to roll back at the first sign of trouble, that would be great.

Somehow https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/workflow/workflow-durable-task-step/1107.ve0b5abf0ca7f/ is dead. Not sure why; https://ci.jenkins.io/job/Plugins/job/workflow-durable-task-step-plugin/job/PR-185/5/console claims to have deployed as the status check confirms. @daniel-beck any clue?

jglick:
CD is also set up here, so if @car-roll merges, then you should have an official release version within the hour.

car-roll:
I can go ahead and merge. Was just waiting for @dduportal 's comments.

jglick:
And there we go @dduportal: https://github.com/jenkinsci/workflow-durable-task-step-plugin/releases/tag/1102.v9c8d2f466adb (should be on UC shortly)

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -1264,6 +1266,46 @@ public void accessPermittedOnlyFromCurrentBuild() throws Throwable {
...
+            WorkflowRun b = p.getBuildByNumber(1);
+            assertFalse(b.isLogUpdated());
+            r.assertBuildStatusSuccess(b);
+            while (Queue.getInstance().getItems().length > 0) {
+                Thread.sleep(100L);
+            }
+            assertThat(logging.getMessages(), hasItem(startsWith("Refusing to build ExecutorStepExecution.PlaceholderTask{runId=p#")));
+        });
+    }
+
```
jglick, line 1308: BTW the reason there was no newline here before is that this step was a `@TestExtension` of the test formerly above it, now with the new test intervening.

dduportal:
> And there we go @dduportal: https://github.com/jenkinsci/workflow-durable-task-step-plugin/releases/tag/1102.v9c8d2f466adb (should be on UC shortly)

Installed, tested and approved: last restart did not triggered "ghosh" builds in the queue. Many thanks for this \o/

dwnusbaum:
@dduportal do you see messages of the form "Refusing to build <...> and cancelling it because associated build is complete" in the Jenkins system logs? What about "Resuming <...>, which is missing from FlowExecutionList (<...>), so registering it now"?

dduportal:
> @dduportal do you see messages of the form "Refusing to build <...> and cancelling it because associated build is complete" in the Jenkins system logs? What about "Resuming <...>, which is missing from FlowExecutionList (<...>), so registering it now"?

Yes, just retried a restart and got the following message:

```
Dec 22, 2021 4:21:02 PM WARNING org.jenkinsci.plugins.workflow.support.steps.ExecutorStepExecution$PlaceholderTask getCauseOfBlockage

Refusing to build ExecutorStepExecution.PlaceholderTask{runId=Plugins/declarative-pipeline-migration-assistant-plugin/PR-117#1,label=maven-windows,context=CpsStepContext[13:null]:Owner[Plugins/declarative-pipeline-migration-assistant-plugin/PR-117/1:Plugins/declarative-pipeline-migration-assistant-plugin/PR-117 #1],cookie=null,auth=null} and cancelling it because associated build is complete
```

dwnusbaum:
Hmm, it's somewhat concerning to me that you are seeing those messages just on a regular restart. Thanks for confirming though!

PR #188, opened 2021-12-15
[JENKINS-67531] Do not load Pipelines during PlaceholderTask.getCauseOfBlockage in normal scenarios

dwnusbaum:
See [JENKINS-67351](https://issues.jenkins.io/browse/JENKINS-67351) and https://github.com/jenkinsci/workflow-api-plugin/pull/188. Amends https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185.

One of the stack traces related to JENKINS-67351 (from the jenkinsci-users mailing list) indicated that a build being accessed in `PlaceholderTask.getCauseOfBlockage` was not yet loaded because `FlowExecutionList$ItemListenerImpl.onLoaded` had not executed yet. I did not think this was possible in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185#discussion_r762246581, and we do not want to be loading Pipelines while the queue lock is held in normal scenarios.

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #197, opened 2022-01-14
Revert fixes related to JENKINS-67351 and JENKINS-67164

dwnusbaum:
Reverts #188 and #185 due to problems in https://github.com/jenkinsci/workflow-api-plugin/pull/188 (see https://github.com/jenkinsci/workflow-api-plugin/pull/188#issuecomment-1013382364) which have no obvious fix and so require the `workflow-api` PRs to be reverted.  https://github.com/jenkinsci/workflow-api-plugin/pull/188 introduced an API required here after #188, but that API will be removed once the workflow-api PRs are reverted, so we need to revert the PRs here as well.

These fixes were generally expected only to apply to uncommon scenarios related to disaster recovery (although they apparently also mattered for regular usage of ci.jenkins.io as mentioned in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/185#issuecomment-999703045), so I think it is ok to revert them for now.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -65,7 +65,6 @@ import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.QueueItemAction;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
 import org.jenkinsci.plugins.workflow.flow.FlowExecution;
-import org.jenkinsci.plugins.workflow.flow.FlowExecutionList;
 import org.jenkinsci.plugins.workflow.flow.FlowExecutionOwner;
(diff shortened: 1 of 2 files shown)


jglick:
> that API will be removed once the workflow-api PRs are reverted

Just deprecate for now?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -424,23 +423,6 @@ public String getCookie() {
         }
 
         @Override public CauseOfBlockage getCauseOfBlockage() {
-            if (FlowExecutionList.get().isResumptionComplete()) {
```
dwnusbaum, line 427: I think I will preserve this API in `workflow-api` temporarily after I revert PRs over there just in case users update `workflow-api` before `workflow-durable-task-step`, but I will just make it always return `false`.

jglick (approved):
:sadpanda:

car-roll (approved):

PR #198, opened 2022-01-20
Support stopOnError for powershell scripts

mbaitelman:
Add support for stopOnError to handle PS errors

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowershellScriptStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowershellScriptStep.java
@@ -37,9 +37,12 @@ public class PowershellScriptStep extends DurableTaskStep {
 
     private final String script;
 
-    @DataBoundConstructor public PowershellScriptStep(String script) {
+    @DataBoundConstructor public PowershellScriptStep(String script, boolean stopOnError) {
         if (script == null) {
             throw new IllegalArgumentException();
+        }
+        if(stopOnError){
+            script = "$ErrorActionPreference=\"Stop\"" + script;
         }
         this.script = script;
     }
--- a/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowerShellStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowerShellStepTest.java
@@ -65,6 +65,13 @@ public class PowerShellStepTest {
         p.setDefinition(new CpsFlowDefinition("node {powershell 'throw \"bogus error\"'}", true));
         j.assertLogContains("bogus error", j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0)));
     }
+
+    // Test that a powershell step that fails indeed causes the underlying build to fail with stopOnFailure: true
+    @Test public void testStopOnFailure() throws Exception {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "stopfailure");
+        p.setDefinition(new CpsFlowDefinition("node {powershell( script: 'throw \"bogus error\"', stopOnFailure: true)}", true));
+        j.assertLogContains("bogus error", j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0)));
+    }
     
     @Test public void testUnicode() throws Exception {
         WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "foobar");


mbaitelman:
I was having trouble with this param until I figured out that this plugin isn't https://github.com/jenkinsci/powershell-plugin which is mentioned everywhere else.

mbaitelman:
Id rather this not be a blocker but if this is accepted the docs will need to be updated as well.
https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#code-powershell-code-powershell-script

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowerShellStepTest.java
```diff
@@ -65,6 +65,13 @@
         p.setDefinition(new CpsFlowDefinition("node {powershell 'throw \"bogus error\"'}", true));
         j.assertLogContains("bogus error", j.assertBuildStatus(Result.FAILURE, p.scheduleBuild2(0)));
     }
+
+    // Test that a powershell step that fails indeed causes the underlying build to fail with stopOnFailure: true
+    @Test public void testStopOnFailure() throws Exception {
+        WorkflowJob p = j.jenkins.createProject(WorkflowJob.class, "stopfailure");
+        p.setDefinition(new CpsFlowDefinition("node {powershell( script: 'throw \"bogus error\"', stopOnFailure: true)}", true));
```
car-roll, line 72: Maybe make the job do more after the error step (like print a line) and verify that the subsequent actions do not appear

car-roll:
Docs can be updated here: https://github.com/jenkinsci/workflow-durable-task-step-plugin/tree/master/src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task

car-roll:
Looks good! I really do think though we need to add some documentation. @dwnusbaum If we are updating the documentation to include this label for just the `Powershell` step, would we need to add a jelly file here: https://github.com/jenkinsci/workflow-durable-task-step-plugin/tree/master/src/main/resources/org/jenkinsci/plugins/workflow/steps/durable_task/PowershellScriptStep ?

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowershellScriptStep.java
```diff
@@ -29,18 +29,27 @@
...
+        this.stopOnError = stopOnError;
+    }
 
     @DataBoundConstructor public PowershellScriptStep(String script) {
         if (script == null) {
             throw new IllegalArgumentException();
         }
+        if(this.stopOnError){
+            script = "$ErrorActionPreference=\"Stop\"" + script;
+        }
```
dwnusbaum, line 52: Does this do anything? The constructor will run before `setStopOnError` could be called, so I think this branch is never entered. That said, if your new tests pass, then maybe the existing behavior is already fine and this PR is not needed, or maybe your tests don't quite cover the right scenario.

Also more generally, I don't think you should modify the script like this here. I think the right way to do this would be to modify [powershellHelper.ps1](https://github.com/jenkinsci/durable-task-plugin/blob/0d250a6142908f591bda91d8658ab1b102611ce3/src/main/resources/org/jenkinsci/plugins/durabletask/powershellHelper.ps1) (or do something like [this](https://github.com/jenkinsci/durable-task-plugin/blob/0d250a6142908f591bda91d8658ab1b102611ce3/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java#L152)) over in `durable-task` plugin and expose a new `@DataBoundSetter` on [`PowershellScript`](https://github.com/jenkinsci/durable-task-plugin/blob/0d250a6142908f591bda91d8658ab1b102611ce3/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java).

dwnusbaum:
@car-roll yeah, if you add new parameters, they need to be added to `config-details.jelly` in that directory, and a `help-parameterName.html` file should be added there.

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/PowershellScriptStep.java
car-roll: Good points. If I had to choose, I'd go with the [second option](https://github.com/jenkinsci/durable-task-plugin/blob/0d250a6142908f591bda91d8658ab1b102611ce3/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java#L152)
`powershellHelper.ps1` does not exist in the [binary version](https://github.com/jenkinsci/lib-durable-task) of `durabletask`, so you would lose support there.

PR #209, opened 2022-03-17
[JENKINS-68080] Temporarily disable push mode tests on Java 17+

basil:
See [JENKINS-68080](https://issues.jenkins.io/browse/JENKINS-68080). I have discovered that the Pipeline: Nodes and Processes tests that enable watch mode consistently hang on Java 17. The default pull mode passes Java 17 tests. To unblock Java 17 testing, this PR skips such tests when running on Java 17. This is a temporary workaround. When JENKINS-68080 is resolved, this workaround can be deleted. I have verified in jenkinsci/bom#935 that with this PR Java 17 tests pass (by skipping the watch mode tests).

PR #212, opened 2022-03-22
[JENKINS-68080] `DurableTaskStep.HandlerImpl.printStreamDelegate` requires `--add-opens`

jglick:
Fixes tests skipped in #209.

https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/bb7ad076d27b281f6213e5571719cc01a729c946/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L697-L702 is the problem

```
hudson.remoting.ProxyException: java.lang.reflect.InaccessibleObjectException: Unable to make field protected java.io.OutputStream java.io.FilterOutputStream.out accessible: module java.base does not "opens java.io" to unnamed module @… [master]
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:354) [master]
	at java.base/java.lang.reflect.AccessibleObject.checkCanSetAccessible(AccessibleObject.java:297) [master]
	at java.base/java.lang.reflect.Field.checkCanSetAccessible(Field.java:178) [master]
	at java.base/java.lang.reflect.Field.setAccessible(Field.java:172) [master]
	at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$HandlerImpl.<clinit>(DurableTaskStep.java:702) [master]
```

`--add-opens java.base/java.io=ALL-UNNAMED` is passed to the controller, but not the agent by [`JenkinsRule.createComputerLauncher`](https://github.com/jenkinsci/jenkins-test-harness/blob/006976316a9ed5c54e72600d4adb01bb33ca24d0/src/main/java/org/jvnet/hudson/test/JenkinsRule.java#L1070-L1074). Normally the error is swallowed because https://github.com/jenkinsci/durable-task-plugin/blob/0d250a6142908f591bda91d8658ab1b102611ce3/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java#L548 runs this asynch and Remoting apparently discards errors from asynch tasks without logging them? (`act` actually works just as well, and exposes the error.)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -699,7 +699,11 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
                 // Defined in Java Platform and protected, so should not happen.
                 throw new ExceptionInInitializerError(x);
             }
-            printStreamDelegate.setAccessible(true);
+            try {
+                printStreamDelegate.setAccessible(true);
+            } catch (/* TODO Java 11+ InaccessibleObjectException */RuntimeException x) {
+                LOGGER.log(Level.WARNING, "On Java 17 error handling is degraded unless `--add-opens java.base/java.io=ALL-UNNAMED` is passed to the agent", x);
+            }
         }
 
         private static final long serialVersionUID = 1L;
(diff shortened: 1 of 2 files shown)


basil (approved):
I have been considering whether to add any `Add-Opens` directives to Remoting's `MANIFEST.MF`. After some discussion in https://github.com/jenkinsci/docker-agent/issues/207#issuecomment-1064513412 @timja and I concluded they may have been needed in the past (perhaps for older versions of XStream?) but do not seem to be needed in the present, so in jenkinsci/docker-inbound-agent#263 I removed them from our Docker images and did not add anything to Remoting's `MANIFEST.MF`. This would be the first case I am aware of where `Add-Opens` is required on the agent side. (On the controller, of course, we know it is needed for XStream.) If this is the only such case, and it is only needed for better error handling, then I suppose it is better to keep Remoting's `MANIFEST.MF` free of any `Add-Opens` entries. If, on the other hand, we find additional use cases, then it might be worth adding `Add-Opens` to Remoting's `MANIFEST.MF`.

jglick:
Ideally we would not need to use reflection to get the wrapped `OutputStream`, but the `TaskListener` interface is built around `PrintStream`. A bunch of code could be made simpler and more robust (and this fixed) if `TaskListener` exposed an optional ability to return the underlying `OutputStream`; I have just not gotten around to it. I agree that adding the JVM flag to agents is probably overkill just to support better error handling in a still-experimental mode.

car-roll (approved):

PR #221, opened 2022-05-08
Make a comment more inclusive

Ladyprowess:
Rather than "master" use "controller"
Rather than "slave" use "agent"

~~Ensure you have provided tests - that demonstrates feature works or fixes the issue~~

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepTest.java
@@ -56,7 +56,7 @@ public class WorkspaceStepTest {
     @Test public void customWorkspaceConcurrency() throws Exception {
         // Currently limited to WorkspaceList.allocate:
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        // Use master as it has 2 executors by default, whereas createSlave hardcodes 1, and I do not want to bother creating a slave by hand:
+        // Use controller as it has 2 executors by default, whereas createSlave hardcodes 1, and I do not want to bother creating a agent by hand:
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/WorkspaceStepTest.java
```diff
@@ -56,7 +56,7 @@ public class WorkspaceStepTest {
     @Test public void customWorkspaceConcurrency() throws Exception {
         // Currently limited to WorkspaceList.allocate:
         WorkflowJob p = r.jenkins.createProject(WorkflowJob.class, "p");
-        // Use master as it has 2 executors by default, whereas createSlave hardcodes 1, and I do not want to bother creating a slave by hand:
+        // Use controller as it has 2 executors by default, whereas createSlave hardcodes 1, and I do not want to bother creating a agent by hand:
```
aHenryJard, line 59: @MarkEWaite I wonder if this one is not the referring to the "master node" and not the "master instance", and should be called "built-in node". "controller" looks good to and is maybe more easy to understand for more plugin developpers.

```suggestion
        // Use the controller as it has 2 executors by default, whereas createSlave hardcodes 1, and I do not want to bother creating an agent by hand:
```

jglick:
Could this be folded into #220 or vice-versa?

MarkEWaite:
> Could this be folded into #220 or vice-versa?

Yes, if that will improve its chances to be merged, then a single pull request that combines both this change and #220 can be created.  Will that improve the changes of it being merged?

MarkEWaite:
> Not exactly necessary, but appreciated.

https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/227 includes both #220 and #221

jglick:
Folded into #227.

PR #224, opened 2022-05-11
Deflake and speed up parts of `ExecutorStepTest`

jtnord:
calling `setLabels` on an agent will not persist the node.

in older versions of Jenkins the tests would be less flaky as adding any
Node would cause all labels to be re-evaluated, so when creating a few
agents and adding labels in a loop the last one created would at least
deterministically ensure that all previous agents labels where correct.

However since 2.332 (https://github.com/jenkinsci/jenkins/pull/5882)
only labels part of a node added or removed would be updated, and when
creating the agents they where created without labels, which where added
later.

This caused tests to be flaky depending on when the periodic
`trimLabels` was called (or at least on other timing related things)

THis was discovered by enabling a loggerrule for hudson.model.queue and
observing that the builds would timeout as not all the agents would have
the expected nodes.

e.g.

```
  12.023 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[ #1] rejected part of demo #1: ?Jenkins? doesn?t have label ?foo?
  12.023 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[slave0 #0] is a potential candidate for task part of demo #1
  12.024 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[slave2 #0] rejected part of demo #1: ?slave2? doesn?t have label ?foo?
  12.024 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[slave1 #0] rejected part of demo #1: ?slave1? doesn?t have label ?foo?
  12.024 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[ #0] rejected part of demo #1: ?Jenkins? doesn?t have label ?foo?
  12.024 [id=141]       FINEST  hudson.model.Queue#maintain: JobOffer[slave3 #0] rejected part of demo #1: ?slave3? doesn?t have label ?foo?
```

from `reuseNodesWithSameLabelsInParallelStages`

Additionally creating agents and waiting for them to come oneline is
slow.  A pipeline will start and then wait for the node to be available,
so we can do other things whilst the agent is connecting.

For the case where we need a number of agents connected before we start
to run the pipeline, we now create iall the agents before waiting for them to connect.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -1196,7 +1196,7 @@ public class ExecutorStepTest {
         // create all the slaves then wait for them to connect as it will be quicker as agents connect in parallel
         ArrayList<DumbSlave> agents = new ArrayList<>();
         for (int i = 0; i < number; ++i) {
-            agents.add(r.createSlave("foo bar", null));
+            agents.add(r.createSlave(label, null));
         }
         for (DumbSlave agent : agents) {
             r.waitOnline(agent);
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
@@ -27,6 +27,7 @@ package org.jenkinsci.plugins.workflow.support.steps;
 import com.gargoylesoftware.htmlunit.Page;
 import com.google.common.base.Predicate;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Functions;
 import hudson.model.Computer;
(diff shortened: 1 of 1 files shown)


jglick:
(could switch label to `developer` if these flakes are holding you up somewhere and you would like an immediate release)

jtnord:
> (could switch label to `developer` if these flakes are holding you up somewhere and you would like an immediate release)

These are currently the most flaky of all, so `developer` would be great (I currently am not a maintainer so can change the label)

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -139,10 +140,8 @@ public class ExecutorStepTest {
      */
     @Test public void buildShellScriptOnSlave() throws Throwable {
         sessions.then(r -> {
-                DumbSlave s = r.createOnlineSlave();
-                s.setLabelString("remote quick");
+                DumbSlave s = r.createSlave("remote quick", null);
```
jglick, line 143: https://javadoc.jenkins.io/component/jenkins-test-harness/org/jvnet/hudson/test/JenkinsRule.html#createSlave-java.lang.String-hudson.EnvVars- for reference

jglick (approved):

PR #226, opened 2022-05-11
Recover gracefully when a `PlaceholderTask` is in the queue but the associated build is complete (II)

jglick:
Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/221. Reapplies #185 (and its amendment #188), reverting the reversion #197.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,6 +95,7 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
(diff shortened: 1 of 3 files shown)


dwnusbaum (approved):
Other than the version in `pom.xml`, this is a straight reversion of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/197 with no new changes, right? Clearer IMO to use `git revert` or cherry-pick the old commits.

jglick:
> no new changes, right?

Right.

> `git revert`

That is what I did, but needed to use `--no-commit` to amend the POM, and so the metadata may not have carried over.

jglick:
https://issues.jenkins.io/browse/JENKINS-51568

PR #241, opened 2022-06-24
Improve `PlaceholderTask.getACL`

jglick:
Slightly less general, but probably more robust, implementation.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -525,8 +525,10 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         public @CheckForNull Run<?,?> runForDisplay() {
             Run<?,?> r = run();
             if (r == null && /* not stored prior to 1.13 */runId != null) {
-                try (ACLContext context = ACL.as(ACL.SYSTEM)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -478,31 +476,24 @@ public String getShortDescription() {
...
-                }
-                Queue.Executable executable = exec.getOwner().getExecutable();
-                if (executable instanceof AccessControlled) {
-                    return ((AccessControlled) executable).getACL();
+            Run<?, ?> r = runForDisplay();
+            if (r != null) {
+                return r.getACL();
+            } else {
+                Jenkins j = Jenkins.get();
+                Job<?, ?> job = j.getItemByFullName(runId.substring(0, runId.lastIndexOf('#')), Job.class);
```
dwnusbaum, line 491: Not sure if it is accurate or relevant here, but `getItemByFullName`'s Javadoc says it can throw `AccessDeniedException`. Maybe we should leave the outer `try/catch` in place just in case?

jglick: Theoretically if you could `DISCOVER` but not `READ` the job, yes. In that case the GUI at least would presumably not be even showing the item (so no red **X** either) but yes we should be defensive here.

dwnusbaum (approved):

PR #242, opened 2022-06-24
Warn when `Queue.cancel` calls fail

jglick:
@dduportal reports a situation where a stale `PlaceholderTask` sits in the queue and apparently cannot be canceled from the GUI (see also #241).


dwnusbaum (approved):

dduportal (approved):
thanks!!

PR #249, opened 2022-07-08
[JENKINS-49707] `AgentErrorCondition` improvements

jglick:
Make `retry(count: …, conditions: [agent()])` handle more cases:

* `Unable to create live FilePath…` which is observed sometimes before an agent is even used for the first time. (https://github.com/jenkins-infra/helpdesk/issues/2984#issuecomment-1178022133)
* The queue item for a `node` block was cancelled. While this _could_ be something that happened due to manual administrator intervention, typically it is a side effect of the cloud killing things off. `KubernetesAgentErrorCondition` (https://github.com/jenkinsci/kubernetes-plugin/pull/1190) already handles this because when a pod is removed, the cloud tried to clean up by removing the agent from the controller’s list of nodes (already handled) _and_ cancelling any outstanding queue items associated with it, but the order is tricky to predict. Note that aborting the overall _build_ is something different.


dduportal (approved):
LGTM functionnaly speaking. Many thanks!

timja (approved):

PR #254, opened 2022-07-21
[JENKINS-49707] `retry` without conditions should handle node removal

jglick:
Downstream of https://github.com/jenkinsci/workflow-step-api-plugin/pull/124, and https://github.com/jenkinsci/workflow-cps-plugin/pull/570 to be effective.

https://github.com/jenkins-infra/helpdesk/issues/3031#issuecomment-1175343746 noted earlier in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/180#discussion_r914058932. If you have

```groovy
retry(2) { // no conditions: […]
  node('remote') {
    sh 'work
  }
}
```

and the agent is removed, the block was surprisingly not retried. This is because `actualInterruption` defaults to true for compatibility in https://github.com/jenkinsci/workflow-step-api-plugin/pull/51 and `BodyExecution.cancel(CauseOfInterruption...)` does not override that.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -287,7 +287,11 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                             continue;
                         }
                         listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
+                        if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
+                            body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -287,7 +287,11 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
                             continue;
                         }
                         listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                        body.cancel(new RemovedNodeCause());
+                        if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
+                            body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
```
jglick, line 291: Passing `false` here is the key aspect of the fix. It means that this was _not_ some user clicking the red **×** on the build, and so `retry` should process it by default (even without use of `conditions: [agent()]`).

dwnusbaum (approved):

car-roll (approved):

jetersen (approved):

PR #287, opened 2023-01-23
Implement `SubTask.getOwnerExecutable`

jglick:
Implements the API introduced in https://github.com/jenkinsci/jenkins/pull/7599, for now without requiring a newer core dep.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -429,6 +429,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             return j.getNode(label);
         }
 
+        @Deprecated
         @Override public boolean isBuildBlocked() {
             return false;
         }
@@ -535,6 +536,10 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             return hasPermission(Item.CANCEL);
         }
 
+        /**
+         * @deprecated use {@link #getOwnerExecutable} (which does not require a dependency on this plugin).
+         */
+        @Deprecated
         public @CheckForNull Run<?,?> run() {
             try {
                 if (!context.isReady()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -429,6 +429,7 @@ public String getCookie() {
             return j.getNode(label);
         }
 
+        @Deprecated
```
jglick, line 432: Not really related, just noticed while I was here.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -721,7 +736,7 @@ private String computeEnclosingLabel(FlowNode executorStepNode, List<FlowNode> h
         }
 
         @Override public long getEstimatedDuration() {
-            Run<?,?> r = run();
+            Run<?,?> r = runForDisplay();
```
jglick, line 739: Also only tangentially related—there was no apparent reason not to be calling the newer and more robust `runForDisplay` in this context.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java
```diff
@@ -613,6 +613,7 @@ private static void assertLogMatches(WorkflowRun build, String regexp) throws IO
                 Queue.Item[] items = Queue.getInstance().getItems();
                 assertEquals(1, items.length);
                 assertEquals(p, items[0].task.getOwnerTask());
+                assertEquals(b, items[0].task.getOwnerExecutable());
```
jglick, line 616: Test is covering something unrelated, but seemed as good a place as any to cover the new method.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -535,6 +536,10 @@ public String getShortDescription() {
             return hasPermission(Item.CANCEL);
         }
 
+        /**
+         * @deprecated use {@link #getOwnerExecutable} (which does not require a dependency on this plugin).
```
daniel-beck, line 540: Potentially confusing without a core bump, as a dependency on this plugin or core ~2.389 is needed.

jglick: True. Will keep this in draft unless & until the core change is merged and I can add a version number.

car-roll (approved):

PR #291, opened 2023-02-15
[JENKINS-70528] Reproducing `FilePathDynamicContext`-related bug in test

jglick:
#270 handled the common case that the nested `node` blocks actually ran on distinct agents. It is however possible (when agents have multiple executors) for the nested `node` to run in a new workspace on the _same_ agent: [JENKINS-70528](https://issues.jenkins.io/browse/JENKINS-70528). Trying to extend the earlier fix to handle this case just breaks `ws` and `dir` steps.


jglick:
waiting on https://github.com/jenkins-infra/helpdesk/issues/3391

jglick:
I have found a cleaner fix. It requires changes in `workflow-basic-steps`, however. I am again considering moving the `ws` & `dir` steps into this plugin to make it easier to work on and test them, since they are subtly coupled to this plugin (https://github.com/jenkinsci/workflow-basic-steps-plugin/pull/86#discussion_r289906309).

PR #292, opened 2023-02-15
[JENKINS-70528] `node` / `dir` / `node` on same agent sets PWD to that of `dir` rather than `@2` workspace

jglick:
[JENKINS-70528](https://issues.jenkins.io/browse/JENKINS-70528)

https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/291#issuecomment-1432225010 subsuming #293

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -80,11 +80,6 @@
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
-            <dependency>
-                <groupId>org.jenkins-ci.plugins.workflow</groupId>
-                <artifactId>workflow-basic-steps</artifactId>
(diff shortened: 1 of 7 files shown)


dwnusbaum (approved):

jglick:
```
jobs\demo\builds\1\log-index: The process cannot access the file because it is being used by another process.
```

Not sure why: https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/563901d6b5fce9bd6911d47db6e0f105e9ab79ee/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepTest.java#L619

PR #299, opened 2023-03-07
[JENKINS-60507] Eventually cancel `node` blocks in anomalous status

jglick:
Observed during a build of https://github.com/jenkinsci/bom/pull/1832. @dduportal et al., any interest in testing this?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -21,6 +21,7 @@ import hudson.model.Item;
 import hudson.model.Job;
 import hudson.model.Label;
 import hudson.model.Node;
+import hudson.model.PeriodicWork;
 import hudson.model.Queue;
 import hudson.model.ResourceList;
 import hudson.model.Result;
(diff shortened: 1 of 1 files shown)


jglick:
Probably too aggressive, trying to tune this a bit. Without knowing how to reproduce the original problem I can only guess.

dduportal:
> Observed during a build of [jenkinsci/bom#1832](https://github.com/jenkinsci/bom/pull/1832). @dduportal et al., any interest in testing this?

Absolutely interested !

dduportal:
@jglick just installed on ci.jenkins:

Capture d’écran 2023-03-24 à 13 30 33


Let's check how ti behaves.
Any particular point to check?

jglick:
> Any particular point to check?

That there are no builds marked aborted improperly. If this works, it should automatically clean up certain cases of builds which were actually corrupted (letting them complete normally if `retry` is used).

jglick:
I guess there have been no obvious problems arising from this? So I think it could be merged.

dduportal:
> I guess there have been no obvious problems arising from this? So I think it could be merged.

Confirmed. Sorry for the delay in answer, but no issue: go go go

car-roll (approved):

PR #302, opened 2023-03-29
deeper the spotsbugs checks

pushker001:
Increase the efficiency of finding a bugs

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -67,6 +67,9 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <jenkins.version>2.361.4</jenkins.version>
+        <spotbugs.effort>Max</spotbugs.effort>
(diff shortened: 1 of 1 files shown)


jglick (changes requested):
Introduces lots of violations. They would need to be either fixed or suppressed.

Also, some of these are not valuable. For example the first one I see

```
Exception is caught when Exception is not thrown in org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.start()
```

is deliberate. Could change the `catch` clause to, e.g., `IOException | InterruptedException | RuntimeException` but this is just more to read and does not improve quality at all.

PR #330, opened 2023-08-09
`build` attribute for `buildProgressBar` should be a `Queue.Executable`, not a `Queue.Task`

dwnusbaum:
See https://github.com/jenkinsci/jenkins/blob/4c5e4f86d1fb3bce5d027446e12259c7edcac27f/core/src/main/resources/lib/hudson/buildProgressBar.jelly#L29.

In practice this did not matter because the Jelly template only ever used `build.url` in this case, which is the same between `PlaceholderTask` and `PlaceholderExecutable`, but it causes problems with some changes I am proposing upstream in https://github.com/jenkinsci/jenkins/pull/8321 where we actually need to find the associated `Run`, see [this comment]( https://github.com/jenkinsci/jenkins/pull/8321/files#r1289340587) in particular.

### Testing done

Tested manually.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -942,7 +942,8 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
          * Occupies {@link Executor} while workflow uses this build agent.
          */
         @ExportedBean
-        private final class PlaceholderExecutable implements ContinuableExecutable, AccessControlled {
+        @Restricted(NoExternalUse.class) // Class must be public for Jelly.
+        public final class PlaceholderExecutable implements ContinuableExecutable, AccessControlled {
 
             @Override public void run() {
                 TaskListener listener = null;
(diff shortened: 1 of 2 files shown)


jglick:
https://issues.jenkins.io/browse/JENKINS-31203 sigh

jglick (approved):
Looks like it should work based on https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/94bdc2b50c6fbab37526804d232eb2f1da971db9/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L1127-L1130

dwnusbaum:
Turns out this does not work on its own; I am investigating.

jglick (approved):

dwnusbaum:
A test failed on Windows, but I think it is unrelated:

```
Expected: is <ABORTED>
     but: was <SUCCESS>
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.jvnet.hudson.test.JenkinsRule.assertBuildStatus(JenkinsRule.java:1450)
	at org.jenkinsci.plugins.workflow.steps.durable_task.ShellStepTest.interruptingAbortsBuildEvenWithReturnStatus(ShellStepTest.java:728)
```

This message in the log is unexpected I think:

```
C:\Jenkins\agent\workspace\_durable-task-step-plugin_PR-330\target\tmp\j h9210212895039935246\workspace\p>ping -n 6 127.0.0.1  1>nul 
Body did not finish within grace period; terminating with extreme prejudice
```

PR #344, opened 2023-09-25
Improve `node` resumption logic in edge cases and allow the executor for `node` to be interrupted in cases where the build is already complete or the execution is corrupted

dwnusbaum:
May address some causes of [JENKINS-51568](https://issues.jenkins.io/browse/JENKINS-51568), and may alleviate some of its symptoms even if it does not prevent the issue.

I have been investigating a few issues related to `node` steps, resumption, and build corruption. This PR helps in at least one case, and it may help in others, but I am filing it now to get an incremental build to test further.

I am trying to fix two primary issues:
1. When a `node` step resumes, I want to make sure we never schedule a second `PlaceholderTask` if one is already in the queue. In addition, I want to try to fix possible race conditions related to `PlaceholderTask.runningTasks` when the `PlaceholderTask` has been serialized in more than one location. More fixes may be needed here.
2. We have seen various cases where the `AsynchronousExecution` for a `node` step for a completed build is somehow still occupying an executor slot. I want to make it so that aborting these executions works correctly even if we do not understand why the `AsynchronousExecution` is outliving the build. 

### Testing done

```[tasklist]
### Submitter checklist
```


pom.xml
```diff
@@ -77,27 +77,11 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.361.x</artifactId>
-                <version>2102.v854b_fec19c92</version>
+                <artifactId>bom-2.387.x</artifactId>
+                <version>2378.v3e03930028f2</version>
```
dwnusbaum, line 81: Picking up https://github.com/jenkinsci/workflow-cps-plugin/pull/759, without which the new test exhibits different problems.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -158,8 +179,8 @@ private static abstract class Translator<T> extends DynamicContext.Typed<T> {
         }
 
         @Override protected FilePath get(DelegatedContext context) throws IOException, InterruptedException {
-            if (LOGGER.isLoggable(Level.FINE)) {
-                LOGGER.fine("ESDC=" + context.get(ExecutorStepDynamicContext.class) + " FPR=" + context.get(FilePathDynamicContext.FilePathRepresentation.class));
+            if (LOGGER.isLoggable(Level.FINER)) {
+                LOGGER.finer("ESDC=" + context.get(ExecutorStepDynamicContext.class) + " FPR=" + context.get(FilePathDynamicContext.FilePathRepresentation.class));
```
dwnusbaum, line 183: Moving this down to `FINER` to make it easier to focus on resumption-related messages with a custom logger.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -428,10 +429,21 @@ public static final class PlaceholderTask implements ContinuedTask, Serializable
...
-                    runningTasks.put(cookie, new RunningTask());
+                    // If Jenkins stops while this step is resuming, there may be a PlaceholderTask in the queue as
+                    // well as in program.dat for the same step. We want to make sure not to create a second task to
+                    // avoid race conditions, so we use putIfAbsent.
+                    // TODO: This helps for runningTasks, but other fields like `stopping` may still be problematic.
+                    // Should we refactor things to guarantee that the relevant state is a singleton? For example,
+                    // introduce a PlaceholderTasksAction that holds a map of a new PlaceholderTaskState class, which
+                    // would hold most of what is currently in PlaceholderTask, and then PlaceholderTask would only
+                    // hold a `String cookie` field and would look up PlaceholderTaskState via the action so it wouldn't
+                    // matter where the task was serialized.
```
dwnusbaum, line 440: IDK if there is a simpler option here. The problem is that in most cases `PlaceholderTask` is only serialized in `program.dat`, but whatever we do needs to be robust against deserialization from both the queue and `program.dat`. Perhaps the current approach is good enough but we could make `ExecutorStepDynamicContext.task` non-final and replace it with any task already in the queue if needed during resumption?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -428,10 +429,21 @@ public static final class PlaceholderTask implements ContinuedTask, Serializable
...
+                    runningTasks.putIfAbsent(cookie, new RunningTask());
                 }
             }
-            LOGGER.log(FINE, "deserializing previously scheduled {0}", this);
+            if (LOGGER.isLoggable(Level.FINE)) {
+                LOGGER.log(FINE, null, new Exception("deserializing previously scheduled " + this));
```
dwnusbaum, line 445: I have seen a case where a `node` step was deserialized 3 times, once from the `Queue` and once from `program.dat`, but IDK about the 3rd location, so I added a stack trace for debugging.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1054,17 +1066,20 @@ public final class PlaceholderExecutable implements ContinuableExecutable, Acces
                             LOGGER.fine(() -> "interrupted " + cookie + " in " + runId);
                             Timer.get().submit(() -> { // JENKINS-46738
                                 Executor thisExecutor = /* AsynchronousExecution. */ getExecutor();
+                                AtomicReference<Boolean> cancelledBodyExecution = new AtomicReference(false);
                                 withExecution(execution -> {
                                     BodyExecution body = execution.body;
                                     if (body != null) {
                                         body.cancel(thisExecutor != null ? thisExecutor.getCausesOfInterruption().toArray(new CauseOfInterruption[0]) : new CauseOfInterruption[0]);
-                                    } else { // anomalous state; perhaps build already aborted but this was left behind; let user manually cancel executor slot
```
dwnusbaum, line 1061: If the execution does not exist, because the build is complete or corrupted or whatever, we never even enter this block and so this logic wasn't reachable.

dwnusbaum:
[`ExecutorStepTest.shellOutputAcrossRestart[watching=false]`](https://ci.jenkins.io/job/Plugins/job/workflow-durable-task-step-plugin/job/PR-344/1/testReport/org.jenkinsci.plugins.workflow.support.steps/ExecutorStepTest/linux_17___Build__linux_17____shellOutputAcrossRestart_watching_false_/) is failing because of the changes to use `putIfAbsent` in `PlaceholderTask.readResolve`. We can probably add a `FlagRule` to clear out the static `PlaceholderTask.runningTasks` field in `ExecutorStepTest` to fix that issue.

jglick:
> clear out the static `PlaceholderTask.runningTasks` field

The proper fix would be to convert this to an instance field of some `@Extension`.

dwnusbaum:
> The proper fix would be to convert this to an instance field of some @Extension.

Yes, although at that point it probably makes sense to go for a larger refactoring to try to avoid problems due to multiple instances of `PlaceholderTask` existing after a restart as well. Maybe the extension could hold both a transient `Map<String, RunningTask> runningTasks` field and a non-transient `Map<String, PlaceholderTaskState> taskStates` field, and then all state in `PlaceholderTask` would be removed and instead accessed indirectly via the extension so it wouldn't matter if there were multiple instances of `PlaceholderTask`.

dwnusbaum:
[923c23d](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/923c23d239542ea8feac2473052f2af411d6c468) is a simple change just to get try to get a clean build.

dwnusbaum:
Hmm, the [new test failed on Windows](https://ci.jenkins.io/job/Plugins/job/workflow-durable-task-step-plugin/job/PR-344/3/testReport/org.jenkinsci.plugins.workflow.support.steps/ExecutorStepRjrTest/windows_11___Build__windows_11____restartDuringStepResumption/). The build was aborted because it took too long for the agent to come back. Looking at the logs I think the problem is that the agent failed to reconnect, but I'm not sure if the problem is that it did not reconnect in time or if it was fundamentally unable to reconnect. I'll try increasing `ExecutorPickle.timeoutForNodeMillis` for the test.

Also though, looking at the logs we should make sure errors in `onResume` clean up outstanding queue entries before the step completes to avoid issues.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -90,12 +91,33 @@ void resume(StepContext context) throws Exception {
...
+                if (item.task instanceof ExecutorStepExecution.PlaceholderTask) {
+                    ExecutorStepExecution.PlaceholderTask itemTask = (ExecutorStepExecution.PlaceholderTask) item.task;
+                    if (task.equals(itemTask)) {
+                        itemRef.set(item);
+                        break;
+                    }
+                }
+            }
+        });
+        // TODO: Do we need to check executor slots too in case there was a task in the queue but it has already started?
```
jglick, line 108: :thinking: could be possible, in case some agents reattached before we got to this point.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -90,12 +91,33 @@ void resume(StepContext context) throws Exception {
...
+                // TODO should also report when !ScheduleResult.created, since that is arguably an error
+                throw new IllegalStateException("queue refused " + task);
+            }
+            LOGGER.fine(() -> "scheduled " + item + " for " + path + " on " + node);
+            itemRef.set(item);
+        } else {
+            LOGGER.fine(() -> "reusing " + itemRef.get() + ", which was already in the queue for " + path + " on " + node);
         }
-        LOGGER.fine(() -> "scheduled " + item + " for " + path + " on " + node);
+        Queue.Item item = itemRef.get();
```
jglick, line 120: BTW it might be more legible to replace the `AtomicReference` idiom with a `Stream` usage, so you do not need two variables.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -158,8 +180,8 @@ private static abstract class Translator<T> extends DynamicContext.Typed<T> {
         }
 
         @Override protected FilePath get(DelegatedContext context) throws IOException, InterruptedException {
-            if (LOGGER.isLoggable(Level.FINE)) {
-                LOGGER.fine("ESDC=" + context.get(ExecutorStepDynamicContext.class) + " FPR=" + context.get(FilePathDynamicContext.FilePathRepresentation.class));
+            if (LOGGER.isLoggable(Level.FINER)) {
```
jglick, line 183: or use a `Supplier` lambda

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -427,11 +433,22 @@ public static final class PlaceholderTask implements ContinuedTask, Serializable
...
-                synchronized (runningTasks) {
-                    runningTasks.put(cookie, new RunningTask());
-                }
+                RunningTasks.get().withRunningTasks(runningTasks -> {
+                    // If Jenkins stops while this step is resuming, there may be a PlaceholderTask in the queue as
+                    // well as in program.dat for the same step. We want to make sure not to create a second task to
+                    // avoid race conditions, so we use putIfAbsent.
+                    // TODO: This helps for runningTasks, but other fields like `stopping` may still be problematic.
+                    // Should we refactor things to guarantee that the relevant state is a singleton? For example,
+                    // introduce a PlaceholderTasksAction that holds a map of a new PlaceholderTaskState class, which
```
jglick, line 442: An action on a `FlowNode` you mean?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: > replace it with any task already in the queue

Why would a task already in queue not be `equals` to this one?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -90,12 +91,33 @@ void resume(StepContext context) throws Exception {
         if (executor != null) {
             throw new IllegalStateException("Already resumed");
         }
-        Queue.Item item = Queue.getInstance().schedule2(task, 0).getItem();
-        if (item == null) {
-            // TODO should also report when !ScheduleResult.created, since that is arguably an error
-            throw new IllegalStateException("queue refused " + task);
+        // If Jenkins restarts or crashes while we are waiting below, on the next startup the task may already be in
+        // the queue, in which case we should reuse it instead of scheduling a second task.
```
jglick, line 95: Hmm, `Queue` scheduling should already reuse items (`ScheduleResult.Existing`) in most cases. You are now implementing `equals` and `hashCode` (see `duplicatesInQueue` logic with `liveGetItems` ) so is this logic still needed?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: They would be equal, but `PlaceholderTask` is not immutable (e.g. `stopping`), so it seems best to make sure that everything is using the same actual object regardless of code path.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: I was thinking on the `Run`, but yes perhaps it could be on the `FlowNode` in which case the action would be simpler at the cost of needing a live execution to find the task, but that might be fine here.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Ah maybe `stopping` can be moved to `PlaceholderExecutable` or `ExecutorStepExecution`?

dwnusbaum: Perhaps, although now that I am looking I am wondering if you can also get weird issues if you restart while the initial `PlaceholderTask` is in the queue before `cookie` and `label` have been set to their final values.

jglick: Right, this should perhaps be a different `Task` class for use after the initial scheduling.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
dwnusbaum: Indeed it can be simplified, see [456010d](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/456010d9c970e0e2ee516d5f07b63dd5425f45ca).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
dwnusbaum: [456010d](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/456010d9c970e0e2ee516d5f07b63dd5425f45ca) removes the need for the reference.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
dwnusbaum: Awkward in this case because `context.get` throws checked exceptions.

jglick: :cry: right, nm

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: It seems like we have a few different ideas of how we could avoid problems here, but for now I am wondering if [42fedcb](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/42fedcbe82f857bee30356e1504700b40c55263c) is good enough.

dwnusbaum:
`ShellStepTest.removingAgentIsFatal[watching=false]` failed on Windows:

> C:\Jenkins\agent\workspace\_durable-task-step-plugin_PR-344\target\tmp\j h1582509129256729822\agent-work-dirs\remote\workspace\p: The process cannot access the file because it is being used by another process.

Looking at the logs I see `Could not connect to remote to send interrupt signal to process`, so I guess the `bat` step is still technically running when the test ends and the behavior seems more or less expected, but then I wonder how the test ever worked on Windows.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
dwnusbaum: Any thoughts on [42fedcb](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/42fedcbe82f857bee30356e1504700b40c55263c)?

dwnusbaum:
Checking whether `ShellStepTest.removingAgentIsFatal[watching=false]` is flaky Windows or consistently fails.

dwnusbaum:
I think there are some potential further improvements that could be done here, but I am not sure those improvements are worth the complexity. Even in its current state I think this PR should improve a few different kinds of issues that users have run into, so I am opening it up for review.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick: Good enough for now.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepRjrTest.java
```diff
@@ -0,0 +1,108 @@
...
+        } finally {
+            iar.stop("custom-label");
+        }
+    }
+
+    private static void restartDuringStepResumption_start(JenkinsRule r) throws Exception {
+        WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node('custom-label') { echo 'node step started'; input('test') }", true));
+        WorkflowRun b = p.scheduleBuild2(0).waitForStart();
+        await().atMost(15, TimeUnit.SECONDS).until(() -> b.getAction(InputAction.class), notNullValue());
```
jglick, line 81: FWIW `RestartableJenkinsRule` + `SemaphoreStep` would likely be easier to work with, at least now that you have fixed `static runningTasks`.

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
dwnusbaum: 😅 https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344#discussion_r1342890151

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepRjrTest.java
dwnusbaum: I think there was some other problem that lead me to use `RealJenkinsRule` but I do not remember exactly. I can check again.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
jglick: Sigh, I keep on forgetting that checked exceptions are basically incompatible with all newer language features.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepRjrTest.java
dwnusbaum: Seems to work fine, although the node takes long enough to reconnect that `TIMEOUT_FOR_NODE_MILLIS` needs to be increased (not sure exactly why). See [10215a6](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/344/commits/10215a67c84fa102ad5d39481a72ab5a06a7d4b7).

jglick (approved):

pom.xml
```diff
@@ -132,6 +116,11 @@
             <artifactId>workflow-basic-steps</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-input-step</artifactId>
+            <scope>test</scope>
+        </dependency>
```
jglick, line 123: Unused after 10215a67c84fa102ad5d39481a72ab5a06a7d4b7?
```suggestion
```

dwnusbaum: Oops yes forgot to remove it.

PR #323, opened 2023-07-07
[JENKINS-52165] Stabilize and better test `USE_WATCHING`

jglick:
[JENKINS-52165](https://issues.jenkins.io/browse/JENKINS-52165): the performance benefits seem important. Have not yet tried to measure impact on (controller) CPU or Remoting channel bandwidth, but can use inotify to verify this allows a long-running `sh` step with copious output to only write to `log` (and maybe `log-index`), whereas in currently default pull mode there are constant writes to `program.dat` (via an atomic temp file) due to https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/6f24fa9eb65bc86c5d34718990f90db968f0003e/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L599-L601 which would otherwise be limited to step transitions.

Downstream of https://github.com/jenkinsci/workflow-api-plugin/pull/294 and https://github.com/jenkinsci/durable-task-plugin/pull/186.

- [x] test failures
- [x] pick up https://github.com/jenkinsci/workflow-api-plugin/pull/290 & https://github.com/jenkinsci/durable-task-plugin/pull/185
- [x] handle #212 more cleanly
- [x] check PCT (https://github.com/jenkinsci/bom/pull/2246 and CloudBees CI)
- [x] log output sometimes lost across restarts (caught in CloudBees CI proprietary tests)
- [x] test against `pipeline-cloudwatch-logs`
- [x] test against `KubernetesPipelineOverridenNamespaceTest`
- [x] parameterize relevant tests to run with and without watching mode
- [X] https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/323#issuecomment-1650618648
- [x] have at least some Windows test coverage
- [x] try https://github.com/jenkins-infra/pipeline-library/pull/707

PR #354, opened 2024-01-26
Simplified `PlaceholderTask`

jglick:
Working on the same test as in #353, I found a (locally unreproducible) case where it appears that `Queue.cancel` from `PlaceholderTask.Callback.finished` triggered `CancelledItemListener`. `PlaceholderTask.stopping` was designed to prevent exactly this; the fact that it did not suggests that there are multiple `PlaceholderTask`s involved. As in #344, the root design flaw is that there is mutable state in `PlaceholderTask` yet it is difficult to ensure Java object identity here. So here I am moving the `stopping` field to transient global state (`RunningTasks`). I am also restoring `cookie` to its original purpose of an actual environment variable value used for nothing more than killing off stray background processes at the end of the block; and using `StepContext` as the primary key whenever we need to look up state. I did consider making `PlaceholderTask` actually immutable but this would require changing the storage of the `label` and `cookie` fields, which is more difficult for two reasons: we need to actually read these values from builds running before the upgrade; and we would need to find some other spot to persist them (such as fields in `ExecutorStepExecution` perhaps). I think these fields are not real problems anyway, since they are only mutated once, in `PlaceholderExecutable.run`, and before that point there does not seem to be a risk of there being multiple copies of the `PlaceholderTask`: the known problems involve deserialization from both the queue and from `ExecutorStepDynamicContext.task`, but `ExecutorStepDynamicContext` is not created until after `cookie` and `label` have been written.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -114,11 +114,10 @@ public class ExecutorPickle extends Pickle {
                     if (task instanceof ExecutorStepExecution.PlaceholderTask) {
                         ExecutorStepExecution.PlaceholderTask placeholder = (ExecutorStepExecution.PlaceholderTask)task;
 
-                        // Non-null cookie means body has begun execution, i.e. we started using a node
                         // PlaceholderTask#getAssignedLabel is set to the Node name when execution starts
                         // Thus we're guaranteeing the execution began and the Node is now unknown.
                         // Theoretically it's safe to simply fail earlier when rehydrating any EphemeralNode... but we're being extra safe.
-                        if (placeholder.getCookie() != null && Jenkins.get().getNode(placeholder.getAssignedLabel().getName()) == null ) {
+                        if (placeholder.hasStarted() && Jenkins.get().getNode(placeholder.getAssignedLabel().getName()) == null ) {
                             if (System.nanoTime() > endTimeNanos) {
                                 Queue.getInstance().cancel(item);
                                 owner.getListener().getLogger().printf("Killed %s after waiting for %s because we assume unknown agent %s is never going to appear%n",
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -69,7 +69,7 @@ public final class ExecutorStepDynamicContext implements Serializable {
 
     private static final long serialVersionUID = 1;
 
-    @NonNull ExecutorStepExecution.PlaceholderTask task;
+    final @NonNull ExecutorStepExecution.PlaceholderTask task;
     final @NonNull String node;
     final @NonNull String path;
     final int depth;
@@ -97,8 +97,6 @@ public final class ExecutorStepDynamicContext implements Serializable {
         if (item == null) {
             throw new IllegalStateException("queue refused " + task);
         }
-        // Try to avoid having distinct a instance of PlaceholderTask here compared to any previously-scheduled task.
-        task = (ExecutorStepExecution.PlaceholderTask)item.task;
         LOGGER.fine(() -> (result.isCreated() ? "scheduled " : " using already-scheduled ") + item + " for " + path + " on " + node);
         TaskListener listener = context.get(TaskListener.class);
         if (!node.isEmpty()) { // unlikely to be any delay for built-in node anyway
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -164,7 +164,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             if (item.task instanceof PlaceholderTask) {
                 PlaceholderTask task = (PlaceholderTask) item.task;
                 if (task.context.equals(context)) {
-                    task.stopping = true;
+                    RunningTasks.run(context, t -> t.stopping = true);
                     if (Queue.getInstance().cancel(item)) {
                         LOGGER.fine(() -> "canceled " + item);
                     } else {
@@ -188,7 +188,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                     if (exec instanceof PlaceholderTask.PlaceholderExecutable) {
                         StepContext actualContext = ((PlaceholderTask.PlaceholderExecutable) exec).getParent().context;
                         if (actualContext.equals(context)) {
-                            PlaceholderTask.finish(((PlaceholderTask.PlaceholderExecutable) exec).getParent().cookie);
+                            PlaceholderTask.finish(context, ((PlaceholderTask.PlaceholderExecutable) exec).getParent().cookie);
                             LOGGER.log(FINE, "canceling {0}", exec);
                             break COMPUTERS;
                         } else {
@@ -251,7 +251,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
             if (li.isCancelled()) {
                 if (li.task instanceof PlaceholderTask) {
                     PlaceholderTask task = (PlaceholderTask) li.task;
-                    if (!task.stopping) {
+                    if (!RunningTasks.get(task.context, t -> t.stopping)) {
                         if (LOGGER.isLoggable(Level.FINE)) {
                             LOGGER.log(Level.FINE, null, new Throwable(li.task + " was cancelled"));
                         }
@@ -395,6 +395,11 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         @Nullable AsynchronousExecution execution;
         /** null until placeholder executable runs */
         @Nullable Launcher launcher;
+        /**
+         * Flag to remember that {@link #stop} is being called, so {@link CancelledItemListener} can be suppressed.
+         * Also used for {@link PlaceholderTask#getCauseOfBlockage}.
+         */
+        boolean stopping;
     }
 
     private static final String COOKIE_VAR = "JENKINS_NODE_COOKIE";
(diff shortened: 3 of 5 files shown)


src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -114,11 +114,10 @@ protected Executor tryResolve() throws Exception {
                     if (task instanceof ExecutorStepExecution.PlaceholderTask) {
                         ExecutorStepExecution.PlaceholderTask placeholder = (ExecutorStepExecution.PlaceholderTask)task;
 
-                        // Non-null cookie means body has begun execution, i.e. we started using a node
                         // PlaceholderTask#getAssignedLabel is set to the Node name when execution starts
                         // Thus we're guaranteeing the execution began and the Node is now unknown.
                         // Theoretically it's safe to simply fail earlier when rehydrating any EphemeralNode... but we're being extra safe.
-                        if (placeholder.getCookie() != null && Jenkins.get().getNode(placeholder.getAssignedLabel().getName()) == null ) {
+                        if (placeholder.hasStarted() && Jenkins.get().getNode(placeholder.getAssignedLabel().getName()) == null ) {
```
jglick, line 120: (just better encapsulation)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
```diff
@@ -97,8 +97,6 @@ void resume(StepContext context) throws Exception {
         if (item == null) {
             throw new IllegalStateException("queue refused " + task);
         }
-        // Try to avoid having distinct a instance of PlaceholderTask here compared to any previously-scheduled task.
-        task = (ExecutorStepExecution.PlaceholderTask)item.task;
```
jglick, line 101: Reverting this bit of #344 which should now be unnecessary: any instances in memory ought to be interchangeable at this point.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -408,19 +413,13 @@ public static final class PlaceholderTask implements ContinuedTask, Serializable
         private final String runId;
         /**
          * Unique cookie set once the task starts.
-         * Serves multiple purposes:
-         * identifies whether we have already invoked the body (since this can be rerun after restart);
```
jglick, line 412: I guess it is still being used in this way (`hasStarted` / `willContinue`) but best leave that detail to those methods.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -434,27 +433,6 @@ public static final class PlaceholderTask implements ContinuedTask, Serializable
             LOGGER.log(FINE, "scheduling {0}", this);
         }
 
-        private Object readResolve() {
```
jglick, line 437: No longer necessary to do anything special here AFAICT.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -875,7 +855,7 @@ private String computeEnclosingLabel(FlowNode executorStepNode, List<FlowNode> h
         }
 
         @Override public String toString() {
-            return "ExecutorStepExecution.PlaceholderTask{runId=" + runId + ",label=" + label + ",context=" + context + ",cookie=" + cookie + ",auth=" + auth + '}';
+            return "ExecutorStepExecution.PlaceholderTask{label=" + label + ",context=" + context + '}';
```
jglick, line 858: Reducing noise and redundancy. `CpsStepContext.toString` already encodes the `Run#externalizableId`; and `cookie` is now rarely interesting. `auth` was never interesting unless you installed `authorized-project`, now deprecated.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -896,34 +876,28 @@ public boolean equals(Object obj) {
             return this.context.equals(other.context);
         }
 
-        private static void finish(@CheckForNull final String cookie) {
-            if (cookie == null) {
+        private static void finish(StepContext context, @CheckForNull final String cookie) {
+            RunningTask runningTask = RunningTasks.remove(context);
+            final AsynchronousExecution execution = runningTask.execution;
```
jglick, line 881: (hide whitespace)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1020,9 +994,6 @@ public final class PlaceholderExecutable implements ContinuableExecutable, Acces
                         env.put("EXECUTOR_NUMBER", String.valueOf(exec.getNumber()));
                         env.put("NODE_LABELS", node.getAssignedLabels().stream().map(Object::toString).collect(Collectors.joining(" ")));
 
-                        RunningTasks.get().withRunningTasks(runningTasks -> {
-                            runningTasks.put(cookie, new RunningTask());
-                        });
```
jglick, line 1025: Just populating it on demand.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1156,9 +1122,7 @@ public Long getTimestamp() {
             }
 
             @Override public boolean willContinue() {
-                return RunningTasks.get().withRunningTasks(runningTasks -> {
-                    return runningTasks.containsKey(cookie);
-                });
+                return hasStarted();
```
jglick, line 1125: I think more comprehensible to just check if `cookie` has been written yet, rather than relying on this map entry having been created.

(Cannot check `label != null` since that would already be true at step start time when running `node('some-label') {…}`.)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1189,8 +1153,6 @@ public String getAbsoluteUrl() {
                 return "PlaceholderExecutable:" + PlaceholderTask.this;
             }
 
-            private static final long serialVersionUID = 1L;
```
jglick, line 1192: Unclear why this was here, when the `PlaceholderExecutable` is not `Serializable`.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1230,20 +1192,31 @@ public Queue.Item itemInQueue() {
...
 
-        synchronized void withRunningTasks(Consumer<Map<String, RunningTask>> fn) {
-            fn.accept(runningTasks);
+        static <T> T get(StepContext context, Function<RunningTask, T> fn) {
+            return fn.apply(find(context));
         }
 
-        static RunningTasks get() {
-            return ExtensionList.lookupSingleton(RunningTasks.class);
+        static void run(StepContext context, Consumer<RunningTask> fn) {
```
jglick, line 1208: More ergonomic access.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -1230,20 +1192,31 @@ public Queue.Item itemInQueue() {
...
+            fn.accept(find(context));
         }
+
+        static RunningTask remove(StepContext context) {
+            RunningTasks holder = ExtensionList.lookupSingleton(RunningTasks.class);
+            synchronized (holder) {
+                RunningTask t = holder.runningTasks.remove(context);
+                return t != null ? t : new RunningTask();
```
dwnusbaum, line 1216: Do we hit the `t == null` condition here in normal cases? I wonder whether it would be worth logging a FINE message in that case for future debugging.

dwnusbaum (approved):

jglick:
Seems to cause some PCT test failures I need to study.

dwnusbaum:
If https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/354/commits/d1a2aa26756457172050bd7707d66fc4730b8a09 fixes test failures in PCT, would it make sense to add a test (or multiple tests) here for whatever it is fixing?

jglick:
Probably yes. The clear problem was the `willContinue` implementation, which caused a test using `mock-slave` to keep an ephemeral agent online after its `node` block ended, and thus to be reused rather than a new agent being created. This seems to have been the cause of more subtle issues in other tests as well. I think I can find a more focused way to reproduce that. (Maybe a test using `OnceRetentionStrategy` needs to be introduced? I do not particularly want to add a test dep on `mock-slave` however.)

While working on that I noticed that `Callback.finished` may not have been running all of its logic consistently, since `finish` had been remove the `RunningTask` entry (which led to a broader reworking of the timing of this entry’s addition and removal). I am not really sure how critical a problem that was: it did not appear to result in any PCT failures.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep2Test.java
```diff
@@ -0,0 +1,180 @@
...
+import org.junit.Test;
+import org.junit.runners.Parameterized;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsSessionRule;
+import org.jvnet.hudson.test.SimpleCommandLauncher;
+import org.jvnet.hudson.test.TestExtension;
+
+/**
+ * Like {@link ExecutorStepTest} but not using {@link Parameterized} which appears incompatible with {@link TestExtension}.
```
jglick, line 62: Did not notice in https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/323 that `PipelineOnlyTaskDispatcher` was not actually getting loaded. (`queueTaskOwnerCorrectWhenRestarting` passed with or without it.)

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep2Test.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    @Test public void cloud() throws Throwable {
+        rr.then(r -> {
+            ExtensionList.lookupSingleton(TestCloud.DescriptorImpl.class);
+            r.jenkins.clouds.add(new TestCloud());
+            var p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("node('test') {}", true));
+            r.assertLogContains("Running on test-1", r.buildAndAssertSuccess(p));
+            r.assertLogContains("Running on test-2", r.buildAndAssertSuccess(p));
```
jglick, line 121: Without d1a2aa26756457172050bd7707d66fc4730b8a09, fails

```
java.lang.AssertionError: 

Expected: a string containing "Running on test-2"
     but: was "Started
[Pipeline] Start of Pipeline
[Pipeline] node
Running on test-1 in …/agents/test-1/workspace/p
[Pipeline] {
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
"
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
	at org.jvnet.hudson.test.JenkinsRule.assertLogContains(JenkinsRule.java:1518)
	at org.jenkinsci.plugins.workflow.support.steps.ExecutorStep2Test.lambda$cloud$2(ExecutorStep2Test.java:121)
```

dwnusbaum (approved):

PR #362, opened 2024-04-01
Fix tests for IPV6 

raul-arabaolaza:
Previous version used a hardcoded version of JTH that prevented transitive updates from plugin-pom, also I found https://github.com/jenkinsci/jenkins-test-harness/pull/712 was incomplete to make the tests on this plugin pass on an IPV6 env, so https://github.com/jenkinsci/jenkins-test-harness/pull/742 was needed also.

I have also updated the core version to get https://issues.jenkins.io/browse/JENKINS-58041

I have not been able to test on windows, I hope the PR builder includes windows support, if not I will have to find someone with a windows machine willing to test this before merge

### Testing done

Run the PCT in a proprietary env with IPV6 support 
`mvn verify` locally

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -66,7 +66,7 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.387.3</jenkins.version>
+        <jenkins.version>2.440.1</jenkins.version>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
@@ -75,8 +75,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.387.x</artifactId>
-                <version>2378.v3e03930028f2</version>
+                <artifactId>bom-2.440.x</artifactId>
+                <version>2907.vcb_35d6f2f7de</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -67,8 +67,6 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <jenkins.version>2.387.3</jenkins.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -66,9 +66,9 @@
     </pluginRepositories>
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
-        <jenkins.version>2.387.3</jenkins.version>
-        <!-- TODO until in plugin-pom -->
-        <jenkins-test-harness.version>2042.v787a_641a_9b_26</jenkins-test-harness.version>
+        <!-- TODO Until in plugin-pom -->
```
jglick, line 69: https://github.com/jenkinsci/plugin-pom/pull/908 needs release apparently. Maybe we should switch to CD.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStepTest.java
```diff
@@ -109,9 +109,9 @@ private static void scriptExitingAcrossRestart1(JenkinsRule r) throws Throwable
         var p = r.createProject(WorkflowJob.class, "p");
         var f = new File(r.jenkins.getRootDir(), "f");
         p.addProperty(new ParametersDefinitionProperty(new StringParameterDefinition("F")));
-        p.setDefinition(new CpsFlowDefinition("node('remote') {if (isUnix()) {sh 'sleep 5 && touch \"$F\"'} else {bat 'ping -n 5 localhost && copy nul \"%F%\" && dir \"%F%\"'}}", true));
+        p.setDefinition(new CpsFlowDefinition("node('remote') {if (isUnix()) {sh 'sleep 5 && touch \"$F\"'} else {bat 'ping -n 5 " + r.getURL().getHost() + " && copy nul \"%F%\" && dir \"%F%\"'}}", true));
```
jglick, line 112: I guess; did you test this on Windows with IPv6?

raul-arabaolaza: No, as I only had access to linux machines with IPv6 so I made sure it will not break any existing pr builder which uses ipv4 on windows and in case it breaks on windows with ipv6 for some reason we can fix later.

PR #372, opened 2024-05-14
Fail fast on cloud node removal

Vlatombe:
This cancels the node block immediately instead of waiting for the node to come in certain conditions when we are sure the node can't come back: using a cloud node, using `OnceRetentionStrategy`.

Creating as draft as I'm experiencing test failures when running the full test class
such as

```
[ERROR] org.jenkinsci.plugins.workflow.support.steps.ExecutorStepDynamicContextTest.onceRetentionStrategyNodeDisappearance -- Time elapsed: 3.218 s <<< FAILURE!
java.lang.AssertionError
        at org.junit.Assert.fail(Assert.java:87)
        at org.junit.Assert.assertTrue(Assert.java:42)
        at org.junit.Assert.assertNotNull(Assert.java:713)
        at org.junit.Assert.assertNotNull(Assert.java:723)
        at org.jenkinsci.plugins.workflow.support.steps.ExecutorStepDynamicContextTest.lambda$onceRetentionStrategyNodeDisappearance$10(ExecutorStepDynamicContextTest.java:241)
        at org.jvnet.hudson.test.JenkinsSessionRule$2.evaluate(JenkinsSessionRule.java:102)
        at org.jvnet.hudson.test.JenkinsRule$1.evaluate(JenkinsRule.java:656)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

https://github.com/Vlatombe/workflow-durable-task-step-plugin/blob/a68e629a9d4aa35184077d549c52b819ce4add23/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java#L240

But I don't exactly get where this comes from

- **Fail fast after cloud node removal**
- **Remove useless assertion**

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -367,7 +367,7 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
                         } else {
                             LOGGER.fine(() -> "rediscovering that " + node + " has been removed and timeout has expired");
                             listener().getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
-                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause());
+                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause(), new ExecutorStepExecution.RemovedNodeTimeoutCause());
                         }
                     }
                     removedNodeDiscovered = 0; // something else; reset
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
@@ -122,7 +122,7 @@ public class ExecutorPickle extends Pickle {
                                 Queue.getInstance().cancel(item);
                                 owner.getListener().getLogger().printf("Killed %s after waiting for %s because we assume unknown agent %s is never going to appear%n",
                                         item.task.getDisplayName(), Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS), placeholder.getAssignedLabel());
-                                throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+                                throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause(), new ExecutorStepExecution.RemovedNodeTimeoutCause());
                             }
                         }
                     }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -26,6 +26,7 @@ import hudson.model.Queue;
 import hudson.model.ResourceList;
 import hudson.model.Result;
 import hudson.model.Run;
+import hudson.model.Slave;
 import hudson.model.TaskListener;
 import hudson.model.TopLevelItem;
 import hudson.model.User;
@@ -38,6 +39,7 @@ import hudson.security.ACL;
 import hudson.security.ACLContext;
 import hudson.security.AccessControlled;
 import hudson.security.Permission;
+import hudson.slaves.AbstractCloudSlave;
 import hudson.slaves.OfflineCause;
 import hudson.slaves.WorkspaceList;
 import java.io.IOException;
@@ -74,6 +76,7 @@ import jenkins.util.Timer;
 import org.acegisecurity.Authentication;
 import org.jenkinsci.plugins.durabletask.executors.ContinuableExecutable;
 import org.jenkinsci.plugins.durabletask.executors.ContinuedTask;
+import org.jenkinsci.plugins.durabletask.executors.OnceRetentionStrategy;
 import org.jenkinsci.plugins.workflow.actions.LabelAction;
 import org.jenkinsci.plugins.workflow.actions.QueueItemAction;
 import org.jenkinsci.plugins.workflow.actions.ThreadNameAction;
@@ -346,40 +349,53 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 return;
             }
             LOGGER.fine(() -> "received node deletion event on " + node.getNodeName());
-            Timer.get().schedule(() -> {
-                Computer c = node.toComputer();
-                if (c == null || c.isOnline()) {
-                    LOGGER.fine(() -> "computer for " + node.getNodeName() + " was missing or online, skipping");
-                    return;
-                }
-                LOGGER.fine(() -> "processing node deletion event on " + node.getNodeName());
-                for (Executor e : c.getExecutors()) {
-                    Queue.Executable exec = e.getCurrentExecutable();
-                    if (exec instanceof PlaceholderTask.PlaceholderExecutable) {
-                        PlaceholderTask task = ((PlaceholderTask.PlaceholderExecutable) exec).getParent();
-                        TaskListener listener;
-                        try {
-                            listener = task.context.get(TaskListener.class);
-                        } catch (Exception x) {
-                            LOGGER.log(Level.WARNING, null, x);
-                            continue;
-                        }
-                        task.withExecution(execution -> {
-                            BodyExecution body = execution.body;
-                            if (body == null) {
-                                listener.getLogger().println("Agent " + node.getNodeName() + " was deleted, but do not have a node body to cancel");
-                                return;
-                            }
-                            listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                            if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
-                                body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
-                            } else { // TODO remove once https://github.com/jenkinsci/workflow-cps-plugin/pull/570 is widely deployed
-                                body.cancel(new RemovedNodeCause());
-                            }
-                        });
+            if (isOneShotAgent(node)) {
+                LOGGER.finest(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
+                cancelOwnerExecution(node, new RemovedNodeCause());
+            } else {
+                LOGGER.finest(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
+                Timer.get().schedule(() -> cancelOwnerExecution(node, new RemovedNodeCause(), new RemovedNodeTimeoutCause()), TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
+            }
+        }
+
+        private static boolean isOneShotAgent(Node node) {
+            return node instanceof AbstractCloudSlave ||
+                    (node instanceof Slave && ((Slave) node).getRetentionStrategy() instanceof OnceRetentionStrategy);
+        }
+
+        private static void cancelOwnerExecution(Node node, CauseOfInterruption... causes) {
+            Computer c = node.toComputer();
+            if (c == null || c.isOnline()) {
+                LOGGER.fine(() -> "computer for " + node.getNodeName() + " was missing or online, skipping");
+                return;
+            }
+            LOGGER.fine(() -> "processing node deletion event on " + node.getNodeName());
+            for (Executor e : c.getExecutors()) {
+                Queue.Executable exec = e.getCurrentExecutable();
+                if (exec instanceof PlaceholderTask.PlaceholderExecutable) {
+                    PlaceholderTask task = ((PlaceholderTask.PlaceholderExecutable) exec).getParent();
+                    TaskListener listener;
+                    try {
+                        listener = task.context.get(TaskListener.class);
+                    } catch (Exception x) {
+                        LOGGER.log(Level.WARNING, null, x);
+                        continue;
                     }
+                    task.withExecution(execution -> {
+                        BodyExecution body = execution.body;
+                        if (body == null) {
+                            listener.getLogger().println("Agent " + node.getNodeName() + " was deleted, but do not have a node body to cancel");
+                            return;
+                        }
+                        listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
+                        if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
+                            body.cancel(new FlowInterruptedException(Result.ABORTED, false, causes));
+                        } else { // TODO remove once https://github.com/jenkinsci/workflow-cps-plugin/pull/570 is widely deployed
+                            body.cancel(causes);
+                        }
+                    });
                 }
-            }, TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
+            }
         }
     }
 
@@ -388,6 +404,12 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         public static boolean ENABLED = Boolean.parseBoolean(System.getProperty(ExecutorStepExecution.class.getName() + ".REMOVED_NODE_DETECTION", "true"));
         @Override public String getShortDescription() {
             return "Agent was removed";
+        }
+    }
+
+    public static final class RemovedNodeTimeoutCause extends CauseOfInterruption {
+        @Override public String getShortDescription() {
+            return "Timeout waiting for agent to come back";
         }
     }
 
--- a/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
+++ b/src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
@@ -119,7 +119,6 @@ public class ExecutorStepDynamicContextTest {
         sessions.then(j -> {
             // Start up a build and then reboot and take the node offline
             assertEquals(0, j.jenkins.getLabel("ghost").getNodes().size()); // Make sure test impl is correctly deleted
-            assertNull(j.jenkins.getNode("ghost")); // Make sure test impl is correctly deleted
             WorkflowRun run = j.jenkins.getItemByFullName("p", WorkflowJob.class).getLastBuild();
             j.assertBuildStatus(Result.ABORTED, j.waitForCompletion(run));
             j.assertLogContains("slave0 has been removed for ", run);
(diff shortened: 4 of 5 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -114,14 +119,17 @@ public class ExecutorStepDynamicContextTest {
         sessions.then(j -> {
             // Start up a build and then reboot and take the node offline
             assertEquals(0, j.jenkins.getLabel("ghost").getNodes().size()); // Make sure test impl is correctly deleted
-            assertNull(j.jenkins.getNode("ghost")); // Make sure test impl is correctly deleted
```
Vlatombe, line 117: Useless assertion, since the node would just have the label `ghost` but usually named `slave0` based on generation.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -346,40 +349,53 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
...
-                                body.cancel(new RemovedNodeCause());
-                            }
-                        });
+            if (isOneShotAgent(node)) {
+                LOGGER.finest(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
+                cancelOwnerExecution(node, new RemovedNodeCause());
+            } else {
+                LOGGER.finest(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
+                Timer.get().schedule(() -> cancelOwnerExecution(node, new RemovedNodeCause(), new RemovedNodeTimeoutCause()), TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
+            }
```
Vlatombe, line 358: This is the key change

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -367,7 +367,7 @@ static final class Execution extends AbstractStepExecutionImpl implements Runnab
                         } else {
                             LOGGER.fine(() -> "rediscovering that " + node + " has been removed and timeout has expired");
                             listener().getLogger().println(node + " has been removed for " + Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS) + ", assuming it is not coming back");
-                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause());
+                            throw new FlowInterruptedException(Result.ABORTED, /* TODO false probably more appropriate */true, new ExecutorStepExecution.RemovedNodeCause(), new ExecutorStepExecution.RemovedNodeTimeoutCause());
```
Vlatombe, line 370: Consistently adding the new `RemovedNodeTimeoutCause` in every place where `TIMEOUT_WAITING_FOR_NODE_MILLIS` timeout has exceeded.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,36 @@ public class ExecutorStepDynamicContextTest {
         });
     }
 
+    @Test public void onceRetentionStrategyNodeDisappearance() throws Throwable {
```
Vlatombe, line 218: Essentially a copy of `normalNodeDisappearance` checking some behavioural differences.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,36 @@ public class ExecutorStepDynamicContextTest {
...
+            WorkflowRun run = j.jenkins.getItemByFullName("p", WorkflowJob.class).getLastBuild();
+            j.assertBuildStatus(Result.ABORTED, j.waitForCompletion(run));
+            j.assertLogNotContains("slave0 has been removed for ", run);
+            assertThat(j.jenkins.getQueue().getItems(), emptyArray());
+            InterruptedBuildAction iba = run.getAction(InterruptedBuildAction.class);
+            assertNotNull(iba);
+            assertThat(iba.getCauses(),
+                    allOf(
+                            contains(isA(ExecutorStepExecution.RemovedNodeCause.class)),
+                            not(contains(isA(ExecutorStepExecution.RemovedNodeTimeoutCause.class)))
```
Vlatombe, line 245: This is the key assertion difference.

Vlatombe:
Looks like some race condition, the `Run` with `InterruptedBuildAction` gets saved to disk properly, but somehow it's not visible in the `Run` object in the next session.

Vlatombe:
Re-launching CI

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
jglick, line 370: Why both? Would it not suffice to use only `RemovedNodeTimeoutCause` (adjusting `AgentErrorCondition` accordingly)?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,31 @@ public class ExecutorStepDynamicContextTest {
...
+        logging.recordPackage(ExecutorStepExecution.class, Level.FINE).record(FlowExecutionList.class, Level.FINE);
+        sessions.then(j -> {
+            // Start up a build that needs executor and then reboot and take the node offline
+            // Starting job first ensures we don't immediately fail if Node comes from a Cloud
+            //  and takes a min to provision
+            WorkflowJob p = j.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("node('ghost') {if (isUnix()) {sh 'sleep infinity'} else {bat 'echo + sleep infinity && ping -n 999999 localhost'}}", true));
+
+            DumbSlave s = j.createSlave(Label.get("ghost"));
+            s.setRetentionStrategy(new OnceRetentionStrategy(0));
```
jglick, line 228: Or you can use `mock-slave` in a test dep, which is somewhat more realistic, though `Cloud` behavior is terribly nondeterministic and hard to control in a test environment.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,31 @@ public class ExecutorStepDynamicContextTest {
         });
     }
 
+    @Test public void onceRetentionStrategyNodeDisappearance() throws Throwable {
+        logging.recordPackage(ExecutorStepExecution.class, Level.FINE).record(FlowExecutionList.class, Level.FINE);
+        sessions.then(j -> {
+            // Start up a build that needs executor and then reboot and take the node offline
+            // Starting job first ensures we don't immediately fail if Node comes from a Cloud
```
jglick, line 222: So should the test actually start the build before creating the agent, and wait for it to wait for an executor?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,31 @@ public class ExecutorStepDynamicContextTest {
...
+            //  and takes a min to provision
+            WorkflowJob p = j.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("node('ghost') {if (isUnix()) {sh 'sleep infinity'} else {bat 'echo + sleep infinity && ping -n 999999 localhost'}}", true));
+
+            DumbSlave s = j.createSlave(Label.get("ghost"));
+            s.setRetentionStrategy(new OnceRetentionStrategy(0));
+            var run = p.scheduleBuild2(0).waitForStart();
+            j.waitForMessage("+ sleep infinity", run);
+            j.jenkins.removeNode(s);
```
jglick, line 231: IIUC this is what `Reaper` in `kubernetes` would do as soon as an agent pod is deleted.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
```diff
@@ -207,4 +215,31 @@ public class ExecutorStepDynamicContextTest {
...
+            j.assertBuildStatus(Result.ABORTED, j.waitForCompletion(run));
+            j.assertLogNotContains("slave0 has been removed for ", run);
```
jglick, line 233: What is this message (not) coming from?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -346,40 +349,53 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
...
-                            }
-                            listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                            if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
-                                body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
-                            } else { // TODO remove once https://github.com/jenkinsci/workflow-cps-plugin/pull/570 is widely deployed
-                                body.cancel(new RemovedNodeCause());
-                            }
-                        });
+            if (isOneShotAgent(node)) {
+                LOGGER.finest(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
```
jglick, line 353:
```suggestion
                LOGGER.fine(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
```
(This is relatively significant.)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -346,40 +349,53 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
...
+                cancelOwnerExecution(node, new RemovedNodeCause());
+            } else {
+                LOGGER.finest(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
```
jglick, line 356:
```suggestion
                LOGGER.fine(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
```

jglick (commented):
I am confused by the doubled-up cause of interruption.

src/main/java/org/jenkinsci/plugins/workflow/support/pickles/ExecutorPickle.java
```diff
@@ -122,7 +122,7 @@ protected Executor tryResolve() throws Exception {
                                 Queue.getInstance().cancel(item);
                                 owner.getListener().getLogger().printf("Killed %s after waiting for %s because we assume unknown agent %s is never going to appear%n",
                                         item.task.getDisplayName(), Util.getTimeSpanString(ExecutorStepExecution.TIMEOUT_WAITING_FOR_NODE_MILLIS), placeholder.getAssignedLabel());
-                                throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause());
+                                throw new FlowInterruptedException(Result.ABORTED, new ExecutorStepExecution.RemovedNodeCause(), new ExecutorStepExecution.RemovedNodeTimeoutCause());
```
jglick, line 125: BTW this pickle is deprecated so there is not really any need to integrate new systems into it.

jglick (commented):
Actually I do not understand why there is a new cause of interruption at all. The only change that should need to be made is for `RemovedNodeListener` to interrupt the build immediately rather than after a delay, right?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
Vlatombe: More realistic, but definitely less deterministic, so I chose my poison.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
Vlatombe: Cleaned that up, it was relevant after restarting an instance, but no longer with the current test scenario.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
Vlatombe: Doesn't matter actually.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextTest.java
Vlatombe: Added an extra test case to cover the cloud node case.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -337,24 +337,29 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
+    public static final class QueueTaskCancelled extends RetryableCauseOfInterruption {
         @Override public String getShortDescription() {
             return Messages.ExecutorStepExecution_queue_task_cancelled();
         }
     }
 
+    /**
+     * A marker interface for {@link CauseOfInterruption} instances that can be retried through {@link AgentErrorCondition}.
+     */
+    public interface Retryable {}
```
jglick, line 349: Maybe clearer to define inside `AgentErrorCondition` then.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -399,20 +404,25 @@ private static void cancelOwnerExecution(Node node, CauseOfInterruption... cause
...
+    public static final class RemovedNodeTimeoutCause extends RetryableCauseOfInterruption {
         @Override public String getShortDescription() {
             return "Timeout waiting for agent to come back";
         }
     }
 
+    /**
+     * Base class for a cause of interruption that can be retried via {@link AgentErrorCondition}.
+      */
+    private abstract static class RetryableCauseOfInterruption extends CauseOfInterruption implements Retryable {}
```
jglick, line 424: Does not seem to contribute anything beyond extending those supertypes directly?

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -334,63 +337,92 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
-                            } else { // TODO remove once https://github.com/jenkinsci/workflow-cps-plugin/pull/570 is widely deployed
-                                body.cancel(new RemovedNodeCause());
-                            }
-                        });
+            if (isOneShotAgent(node)) {
+                LOGGER.fine(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
+                cancelOwnerExecution(node, new RemovedNodeCause());
+            } else {
+                LOGGER.fine(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
+                Timer.get().schedule(() -> cancelOwnerExecution(node, new RemovedNodeTimeoutCause()), TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
```
jglick, line 362: I think
```suggestion
                Timer.get().schedule(() -> cancelOwnerExecution(node, new RemovedNodeCause()), TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
```
would still be clearer? The agent was removed and that is why the build is being aborted, regardless of the fact that this was a static agent and so we offered a 5m grace period for it to be redefined and reconnected (which expired).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -334,63 +337,92 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
-                            listener.getLogger().println("Agent " + node.getNodeName() + " was deleted; cancelling node body");
-                            if (Util.isOverridden(BodyExecution.class, body.getClass(), "cancel", Throwable.class)) {
-                                body.cancel(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
-                            } else { // TODO remove once https://github.com/jenkinsci/workflow-cps-plugin/pull/570 is widely deployed
-                                body.cancel(new RemovedNodeCause());
-                            }
-                        });
+            if (isOneShotAgent(node)) {
+                LOGGER.fine(() -> "Cancelling owner run for one-shot agent " + node.getNodeName() + " immediately");
+                cancelOwnerExecution(node, new RemovedNodeCause());
```
jglick, line 359: (The crucial part FTR.)

jglick (approved):
Simpler now, thanks. Some optional suggestions.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
Vlatombe: Yes, this is just to avoid boilerplate. Can inline just as well.

jglick (approved):

jglick:
`ShellStepTest.removingAgentIsFatal` needs a change to expected message.

jglick (approved):

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -346,51 +349,75 @@ public static final class QueueTaskCancelled extends CauseOfInterruption {
...
+            } else {
+                LOGGER.fine(() -> "Will cancel owner run for agent " + node.getNodeName() + " after waiting for " + TIMEOUT_WAITING_FOR_NODE_MILLIS + "ms");
+                Timer.get().schedule(() -> cancelOwnerExecution(node, new RemovedNodeCause()), TIMEOUT_WAITING_FOR_NODE_MILLIS, TimeUnit.MILLISECONDS);
+            }
+        }
+
+        private static boolean isOneShotAgent(Node node) {
+            return node instanceof AbstractCloudSlave ||
+                    (node instanceof Slave && ((Slave) node).getRetentionStrategy() instanceof OnceRetentionStrategy);
```
jglick, line 363: Unfortunately this heuristic does not match `EC2AbstractSlave extends Slave` nor `EC2RetentionStrategy extends RetentionStrategy`. I guess we need to hard-code support for those nonstandard implementations. CC @car-roll

Vlatombe: Can we start introducing a marker interface for this usage?

car-roll: https://github.com/jenkinsci/ec2-plugin/pull/998

PR #382, opened 2024-07-05
Parallel build can fail to resume when some nodes specify a label which is also a self-label of another node

jglick:


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep2Test.java
```diff
@@ -185,4 +189,35 @@ private TestCloudSlave(String name) throws Exception {
...
 
+    @Test public void selfNameVsLabel() throws Throwable {
+        logging.recordPackage(ExecutorStepExecution.class, Level.FINE);
+        rr.then(r -> {
+            r.jenkins.setNumExecutors(0);
+            ExtensionList.lookupSingleton(DelayX.class).active = false;
+            r.createSlave("x", null, null);
+            r.createSlave("x2", "x", null);
+            WorkflowJob p = r.createProject(WorkflowJob.class, "p");
+            p.setDefinition(new CpsFlowDefinition("parallel b1: {node('x && !x2') {semaphore 'b1'}}, b2: {node('x2') {semaphore 'b2'}}", true));
```
jglick, line 200: More naturally just
```suggestion
            p.setDefinition(new CpsFlowDefinition("parallel b1: {node('x') {semaphore 'b1'}}, b2: {node('x') {semaphore 'b2'}}", true));
```
but I am trying to make the test (without fix) fail more reliably.

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStep2Test.java
```diff
@@ -185,4 +189,35 @@ private TestCloudSlave(String name) throws Exception {
...
+            SemaphoreStep.success("b2/1", null);
+            r.waitForCompletion(b);
+            r.assertBuildStatusSuccess(b);
+        });
+    }
+    @TestExtension("selfNameVsLabel") public static final class DelayX extends ComputerListener {
+        boolean active = true;
+        @Override public void preOnline(Computer c, Channel channel, FilePath root, TaskListener listener) throws IOException, InterruptedException {
+            if (c.getName().equals("x")) {
+                Thread.sleep(5_000);
```
jglick, line 218: Ditto; trying to make `x2` come back up first so that the task originally run on `x` tries to get scheduled on it.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -539,6 +542,27 @@ public boolean hasStarted() {
             return j.getNode(label);
```
jglick, line 542: Note that this method ^^^ _looks_ like it would have taken care of the bug, but in fact is dead code: https://github.com/jenkinsci/jenkins/pull/9442

dwnusbaum (approved):

Vlatombe (commented):
👍

PR #405, opened 2024-11-05
Extending `AnomalousStatus` to also kill `sh` steps

jglick:
In many cases when restoring a build into another K8s cluster using a very lossy backup of the filesystem, via EFS Replication (which does not guarantee snapshot semantics), there is some sort of problem with metadata, which prevents `node` block retry from recovering automatically. (Prior to https://github.com/jenkinsci/kubernetes-plugin/pull/1617 it did not work even if metadata was perfect.) Sometimes there is a missing `program.dat`, sometimes a corrupted log file, sometimes a missing `FlowNode`, etc.

But in many of these cases ([CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-53259)), the log seems fine and the flow nodes seem fine, yet for reasons I cannot easily follow because `program.dat` is so opaque, the `node` block seems to have received an `Outcome.abnormal` with the expected `FlowInterrupedException` from `ExecutorStepDynamicContext.resume`; there are also some suppressed exceptions, and `AnomalousStatus` just adds to this list, without causing the build to proceed. Calling `CpsStepContext.scheduleNextRun()` from the script console does not help either. However in most of these cases it _does_ seem to work to abort the `sh` step running inside: somehow that “wakes up” the program, which then fails the `node` block in the expected way, letting the `retry` step kick in and ultimately letting the build run to completion.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -312,6 +312,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 }
             }
             Set<StepContext> newAnomalous = new HashSet<>();
+            Set<String> affectedNodes = new HashSet<>();
             StepExecution.applyAll(ExecutorStepExecution.class, exec -> {
                 StepContext ctx = exec.getContext();
                 if (!knownTasks.contains(ctx)) {
@@ -323,19 +324,8 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                             LOGGER.log(Level.WARNING, null, x);
                         }
                         ctx.onFailure(new FlowInterruptedException(Result.ABORTED, false, new QueueTaskCancelled()));
-                        // Also abort any shell steps running on the same node:
                         if (exec.state != null) {
-                            try {
-                                for (var exec2 : ctx.get(FlowExecution.class).getCurrentExecutions(true).get()) {
-                                    if (exec2 instanceof DurableTaskStep.Execution) {
-                                        if (exec.state.node.equals(((DurableTaskStep.Execution) exec2).node)) {
-                                            exec2.getContext().onFailure(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
-                                        }
-                                    }
-                                }
-                            } catch (Exception x) {
-                                LOGGER.log(Level.WARNING, null, x);
-                            }
+                            affectedNodes.add(exec.state.node);
                         }
                     } else {
                         newAnomalous.add(ctx);
@@ -345,6 +335,21 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                 }
                 return null;
             }).get();
+            // Also abort any shell steps running on the same node(s):
+            if (!affectedNodes.isEmpty()) {
+                StepExecution.applyAll(DurableTaskStep.Execution.class, exec -> {
+                    if (affectedNodes.contains(exec.node)) {
+                        StepContext ctx = exec.getContext();
+                        try {
+                            ctx.get(TaskListener.class).error("also cancelling shell steps running on " + exec.node);
+                        } catch (IOException | InterruptedException x) {
+                            LOGGER.log(Level.WARNING, null, x);
+                        }
+                        ctx.onFailure(new FlowInterruptedException(Result.ABORTED, false, new RemovedNodeCause()));
+                    }
+                    return null;
+                });
+            }
             for (StepContext ctx : newAnomalous) {
                 ctx.get(TaskListener.class).error("node block appears to be neither running nor scheduled; will cancel if this condition persists");
             }
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -90,6 +90,7 @@ import org.jenkinsci.plugins.workflow.steps.BodyExecutionCallback;
 import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 import org.jenkinsci.plugins.workflow.steps.StepContext;
 import org.jenkinsci.plugins.workflow.steps.StepExecution;
+import org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep;
 import org.jenkinsci.plugins.workflow.steps.durable_task.Messages;
 import org.jenkinsci.plugins.workflow.support.actions.WorkspaceActionImpl;
 import org.jenkinsci.plugins.workflow.support.concurrent.Timeout;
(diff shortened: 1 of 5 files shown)


jglick:
Ineffective.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -322,6 +323,20 @@ public void stop(@NonNull Throwable cause) throws Exception {
                             LOGGER.log(Level.WARNING, null, x);
                         }
                         ctx.onFailure(new FlowInterruptedException(Result.ABORTED, false, new QueueTaskCancelled()));
+                        // Also abort any shell steps running on the same node:
+                        if (exec.state != null) {
+                            try {
+                                for (var exec2 : ctx.get(FlowExecution.class).getCurrentExecutions(true).get()) {
```
jglick, line 329: This just hangs.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -330,6 +335,21 @@ public void stop(@NonNull Throwable cause) throws Exception {
                 }
                 return null;
             }).get();
+            // Also abort any shell steps running on the same node(s):
+            if (!affectedNodes.isEmpty()) {
+                StepExecution.applyAll(DurableTaskStep.Execution.class, exec -> {
+                    if (affectedNodes.contains(exec.node)) {
```
dwnusbaum, line 341: Are there cases (e.g. nodes with multiple executors?) where this could abort steps which are running fine if another step on the same node is having unusual problems? If so, could we check `exec.state.cookie` or something more precise than just the node name instead?

dwnusbaum (approved):

Spanish93:
Hi @jglick,

This merge is causing significant issues. When there is a brief disconnection between the master and the slave, I now encounter the error: "also cancelling shell steps."

As a workaround, I had to revert to version 371.vb_7cec8f3b_95e, which works fine. However, I'm currently blocked on this version due to this problem.

Could you please provide guidance or suggest a solution to resolve this issue?

jglick:
@Spanish93 I suspect something else is wrong in your system. A temporary disconnection of an agent should not trigger this logic; https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/4043ebfb57520bf9a257bbca47355d86bb76ae2b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L308 and https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/4043ebfb57520bf9a257bbca47355d86bb76ae2b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L334 should be logged if you have the corresponding logger enabled at a fine level. Prior to this PR, if this periodic task detected an issue—typically caused either by a queue item being lost due to a crash before any agent even began running a `node` block (since corrected in #408), or some sort of unclassified corruption of build metadata—then the build would still have been aborted: https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/4043ebfb57520bf9a257bbca47355d86bb76ae2b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L322-L326 Perhaps your problem actually already existed, but was less frequently observed due to the longer recurrence period between checks? Do you have any means of reliably reproducing your problem in a self-contained environment?

Spanish93:
Hi @jglick,

After extensive testing, I can confirm that the update is causing issues. When I update the plugin, the job consistently fails. However, reverting to the previous version resolves the problem, and the job completes successfully every time.

jglick:
@Spanish93 exactly how long is

> a brief disconnection

? This plugin already included code to terminate `sh` steps after (by default) 5m of being unable to contact the agent.

And @Sunny-Anand 

> the steps are executed on separate nodes and when one node finishes , the other node where the other step is in progress gets Terminated

similarly, there is no reason that I can see for the change in this PR to cause such a bug unless there was _already_ a problem in your environment, since it is only

* reducing the amount of time between successive runs of an existing check
* making a failed check, that already printed an error to the build log and sent an interrupt to the `node` step which would normally abort the build, also send an interrupt to the corresponding `sh` step

Again, if someone has _steps to reproduce a problem from scratch_ I will do my best to investigate. Enabling `FINE` logging may also offer some clues.

Spanish93:
Hi @jglick ,

Less than 1 second.

jglick:
How I tried to reproduce:
1. made a temp dir `/tmp/jdh` with `a+w`
2. `docker run --rm --name jenkins -p 8080:8080 -v /tmp/djh:/var/jenkins_home jenkins/jenkins:lts`
3. logged in with initial admin password, selected no plugins
4. installed `workflow-durable-task-step`, `workflow-basic-steps`, `workflow-job`, `workflow-cps`
5. created a job
```groovy
parallel a: {
    node('agent1') {
        sh 'set +x; for x in `seq 00 99`; do echo a$x; sleep 6; done'
    }
}, b: {
    node('agent2') {
        sh 'set +x; for x in `seq 00 99`; do echo b$x; sleep 6; done'
    }
}
```
6. created an `agent1` with remote FS `/home/jenkins/agent` using inbound launch method; created `agent2` as a clone of it
7. launched `docker run --rm --init --link jenkins jenkins/inbound-agent -url http://jenkins:8080/ -name agent1 -webSocket -secret …` and similarly for `agent2`
8. triggered a build, watched console until it got up to about `a8` / `b8`
9. `docker network disconnect bridge <name-of-agent1-container>`
10. waited until a few `b…` messages were printed without corresponding `a…` messages; also saw `Cannot contact agent1: java.lang.InterruptedException`
11. `docker network connect bridge <name-of-agent1-container>`
12. batched-up `a…` messages printed as expected, and both branches resumed printing
13. waited a few more minutes, and build completed normally

Not sure if @Spanish93 & @Sunny-Anand are discussing some fundamentally different scenario. The whole `AnomalousStatus` extension can of course be disabled, but I am guessing something is fundamentally wrong in these controllers and that would just be masking the issue.

Spanish93:
Hi @jglick ,

Any news about this big issue ?
Because more and more plugins depends on the latest version of this plugin but because of the problem I can't update this plugin and much other now depending on this.

jglick:
> Any news

No, I am waiting for more information from people claiming to experience the issue.

Spanish93:
Hi @jglick ,

## Issue: Incorrect Shell Step Cancellation Behavior

### Problem Description

The code has two cancellation phases:

1. **First phase** - Cancel anomalous `ExecutorStepExecution` contexts:
   ```java
   ctx.get(TaskListener.class).error("node block still appears to be neither running nor scheduled; cancelling");
   ```

2. **Second phase** - Cancel all shell steps on affected nodes:
   ```java
   ctx.get(TaskListener.class).error("also cancelling shell steps running on " + exec.node);
   ```

### Observed Behavior

- I **do not see** the first message in Jenkins logs
- I **do see** the second message, followed by shell steps being killed

### Root Cause Analysis

The first message likely fails silently because the `ExecutorStepExecution` context is in a "dead" state where:
- The context still exists in memory but the `TaskListener` is no longer functional
- The `IOException`/`InterruptedException` is caught and logged at WARNING level, but the process continues
- The node gets added to `affectedNodes` even though the original error message was never displayed

However, the shell step contexts (`DurableTaskStep.Execution`) are still "alive" and connected to the build log, so their cancellation messages appear successfully.

### Impact

This creates a **critical issue** where:
1. The original problematic executor step fails to show its cancellation message
2. But all shell steps running on the same node get cancelled as "collateral damage"
3. **This kills legitimate, correctly running jobs** on the affected nodes

### Proposed Solution

The automatic shell step cancellation should be conditional or removed entirely, as it's causing more harm than good by killing healthy jobs when only the executor allocation was problematic.

Could we either:
1. Remove the automatic shell step cancellation entirely
2. Add better validation before adding nodes to `affectedNodes`
3. Only cancel shell steps if we can confirm the first cancellation message was actually delivered

This behavior is blocking our pipeline operations as it's terminating valid running jobs.

Thanks for looking into this!

jglick:
> The context still exists in memory but the `TaskListener` is no longer functional

Then you would not see either message in the build log. Perhaps I am misunderstanding your description. (BTW this is not a root cause analysis; the root cause would be whatever breaks the state of the `ExecutorStepExecution` to begin with.)

jglick:
You can test fixes to most plugin PRs without waiting for them to even be released: just wait for the build to pass CI (green) and then look at the **Incrementals** check in the PR for details.

Spanish93:
HI @jglick,

After some days with the fixed version all seem to work well now.

Thanks

PR #408, opened 2024-11-20
Build fails to resume if controller crashes before queue is saved

Vlatombe:
This covers a case where a queue item for a node was filed, then the controller shut down without going through clean up.
Because of that, the queue wasn't saved, and when the controller restarts, the queue item is gone.

This modifies the behaviour of `ExecutorStepExecution#onResume` to detect this condition and re-schedule a new queue item if needed.

### Testing done

See provided unit test.

### Submitter checklist

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,8 @@
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO Until in plugin-pom -->
         <jenkins-test-harness.version>2357.vf2a_982b_b_910f</jenkins-test-harness.version>
-        <jenkins.version>2.479.1</jenkins.version>
+        <jenkins.baseline>2.479</jenkins.baseline>
+        <jenkins.version>${jenkins.baseline}.1</jenkins.version>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
@@ -77,7 +78,7 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.479.x</artifactId>
+                <artifactId>bom-${jenkins.baseline}.x</artifactId>
                 <version>3613.v584fca_12cf5c</version>
                 <scope>import</scope>
                 <type>pom</type>
--- a/pom.xml
+++ b/pom.xml
@@ -67,7 +67,7 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO Until in plugin-pom -->
-        <jenkins-test-harness.version>2182.v0138ccb_c0b_cb_</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2357.vf2a_982b_b_910f</jenkins-test-harness.version>
         <jenkins.version>2.440.1</jenkins.version>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
@@ -114,6 +114,11 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-basic-steps</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>pipeline-input-step</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -78,7 +78,7 @@
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
                 <artifactId>bom-2.479.x</artifactId>
-                <version>3696.vb_b_4e2d1a_0542</version>
+                <version>3613.v584fca_12cf5c</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
                 <artifactId>bom-${jenkins.baseline}.x</artifactId>
-                <version>3613.v584fca_12cf5c</version>
+                <version>3482.vc10d4f6da_28a_</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/pom.xml
+++ b/pom.xml
@@ -28,7 +28,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.88</version>
+        <version>5.3</version>
         <relativePath/>
     </parent>
     <groupId>org.jenkins-ci.plugins.workflow</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -68,7 +68,7 @@
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO Until in plugin-pom -->
         <jenkins-test-harness.version>2357.vf2a_982b_b_910f</jenkins-test-harness.version>
-        <jenkins.version>2.440.1</jenkins.version>
+        <jenkins.version>2.479.1</jenkins.version>
         <useBeta>true</useBeta>
         <gitHubRepo>jenkinsci/${project.artifactId}-plugin</gitHubRepo>
         <hpi.compatibleSinceVersion>2.40</hpi.compatibleSinceVersion>
@@ -77,8 +77,8 @@
         <dependencies>
             <dependency>
                 <groupId>io.jenkins.tools.bom</groupId>
-                <artifactId>bom-2.440.x</artifactId>
-                <version>2907.vcb_35d6f2f7de</version>
+                <artifactId>bom-2.479.x</artifactId>
+                <version>3696.vb_b_4e2d1a_0542</version>
                 <scope>import</scope>
                 <type>pom</type>
             </dependency>
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -216,7 +216,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
         try {
             if (state == null) {
                 var flowNode = getContext().get(FlowNode.class);
-                LOGGER.fine(() -> "No ExecutorStepDynamicContext found for node block " + getContext() + "; will attempt to recover");
+                LOGGER.fine(() -> "node block " + getContext() + " not yet scheduled, checking for an existing queue item");
                 if (flowNode == null) {
                     LOGGER.fine(() -> "No FlowNode found for node block " + getContext() + "; can't recover" );
                 } else {
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
@@ -225,7 +225,7 @@ public class ExecutorStepExecution extends AbstractStepExecutionImpl {
                         LOGGER.fine(() -> "No QueueItemAction found for node block " + getContext() + "; can't recover");
                     } else {
                         LOGGER.fine(() -> "QueueItemAction with id=" + action.id + " found for node block " + getContext());
-                        var queueItem = Queue.getInstance().getItem(action.id);
+                        var queueItem = action.itemInQueue();
                         if (queueItem == null) {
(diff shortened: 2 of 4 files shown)


src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextRJRTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+    @Rule
+    public InboundAgentRule iar = new InboundAgentRule();
+
+    @Test public void restartWhileWaitingForANode() throws Throwable {
+        rjr.startJenkins();
+        try (var tailLog = new TailLog(rjr, "p", 1)) {
+            iar.createAgent(rjr, InboundAgentRule.Options.newBuilder().name("J").label("mib").color(PrefixedOutputStream.Color.YELLOW).webSocket().build());
+            rjr.runRemotely(ExecutorStepDynamicContextRJRTest::setupJobAndStart);
+            rjr.stopJenkinsForcibly();
```
Vlatombe, line 69: The same test passes with
```
rjr.stopJenkins();
```

pom.xml
```diff
@@ -67,7 +67,7 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO Until in plugin-pom -->
-        <jenkins-test-harness.version>2182.v0138ccb_c0b_cb_</jenkins-test-harness.version>
+        <jenkins-test-harness.version>2357.vf2a_982b_b_910f</jenkins-test-harness.version>
```
Vlatombe, line 70: Need https://github.com/jenkinsci/jenkins-test-harness/pull/876

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextRJRTest.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private static void setupJobAndStart(JenkinsRule r) throws Exception {
+        var p = r.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("parallel 'Branch 1': { node('mib') { input 'Complete branch 1 ?' } }, 'Branch 2': { sleep 1; node('mib') { input 'Complete branch 2 ?' } }", true));
+        var b = p.scheduleBuild2(0).waitForStart();
+        r.waitForMessage("Complete branch 1 ?", b);
+        assertTrue(b.isBuilding());
+        await().until(() -> r.jenkins.getQueue().getItems(), arrayWithSize(1));
```
jglick, line 125: Looks like a race condition—will the test still run OK if branch 2 gets as far as starting `node` by this point?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContextRJRTest.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+
+    private static void proceed(List<InputAction> actions, String inputMessage, String name) throws InterruptedException, TimeoutException, IOException {
+        for (var action : actions) {
+            if (action.getExecutions().isEmpty()) {
+                continue;
+            }
+            var inputStepExecution = action.getExecutions().get(0);
+            if (inputMessage.equals(inputStepExecution.getDisplayName())) {
```
jglick, line 111: Clearer to use an `id`. (Just beware that they are forcibly capitalized, so use a capitalized id to start with.)

Vlatombe:
The given test sometimes succeeds...

Vlatombe:
The case where it succeeds is when the stop happens before entering the second node step (still in `sleep`).

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -215,8 +215,25 @@ public void stop(@NonNull Throwable cause) throws Exception {
     @Override public void onResume() {
         try {
             if (state == null) {
-                Run<?, ?> run = getContext().get(Run.class);
-                LOGGER.fine(() -> "No ExecutorStepDynamicContext found for node block in " + run + "; perhaps loading from a historical build record, hoping for the best");
+                var flowNode = getContext().get(FlowNode.class);
+                LOGGER.fine(() -> "No ExecutorStepDynamicContext found for node block " + getContext() + "; will attempt to recover");
```
jglick, line 219: So the previous message misleadingly implied that this could only happen with a build record created by an old plugin version, which is not the case—this is a normal situation. IOW
```suggestion
                LOGGER.fine(() -> "node block " + getContext() + " not yet scheduled, checking for an existing queue item");
```
would be clearer.

pom.xml
```diff
@@ -67,8 +67,8 @@
     <properties>
         <changelist>999999-SNAPSHOT</changelist>
         <!-- TODO Until in plugin-pom -->
-        <jenkins-test-harness.version>2182.v0138ccb_c0b_cb_</jenkins-test-harness.version>
-        <jenkins.version>2.440.1</jenkins.version>
+        <jenkins-test-harness.version>2357.vf2a_982b_b_910f</jenkins-test-harness.version>
+        <jenkins.version>2.479.1</jenkins.version>
```
jglick, line 71: BTW https://github.com/jenkinsci/archetypes/pull/737 while we are here

jglick:
(For reference, this is a corner case which would previously trigger `AnomalousStatus`, as most recently touched in #405.)

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -215,8 +215,25 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
         try {
             if (state == null) {
-                Run<?, ?> run = getContext().get(Run.class);
-                LOGGER.fine(() -> "No ExecutorStepDynamicContext found for node block in " + run + "; perhaps loading from a historical build record, hoping for the best");
+                var flowNode = getContext().get(FlowNode.class);
+                LOGGER.fine(() -> "node block " + getContext() + " not yet scheduled, checking for an existing queue item");
+                if (flowNode == null) {
+                    LOGGER.fine(() -> "No FlowNode found for node block " + getContext() + "; can't recover" );
+                } else {
+                    var action = flowNode.getAction(QueueItemActionImpl.class);
```
jglick, line 223: To remind myself: this would have been added right away in https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/6ae8123a901592f1c4ce1920d4968826a2d8d665/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L132-L139

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
```diff
@@ -215,8 +215,25 @@ public void stop(@NonNull Throwable cause) throws Exception {
...
+                    if (action == null) {
+                        LOGGER.fine(() -> "No QueueItemAction found for node block " + getContext() + "; can't recover");
+                    } else {
+                        LOGGER.fine(() -> "QueueItemAction with id=" + action.id + " found for node block " + getContext());
+                        var queueItem = action.itemInQueue();
+                        if (queueItem == null) {
+                            LOGGER.fine(() -> "Could not find queue item " + action.id + ", rescheduling it");
+                            start();
```
jglick, line 231: I thought of a minor problem with this: https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/6ae8123a901592f1c4ce1920d4968826a2d8d665/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L139 will add a _second_ `QueueItemActionImpl`, which means that if the item is still waiting in queue when the controller next terminates (even gracefully), the next call to `onResume` will find the first action in https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/711657d1bde3b1acca492f53367dbe5441ebb4e4/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L223 and so it will schedule a duplicate queue item.

src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java
jglick, line 223: or
```suggestion
                    var action = flowNode.getPersistentAction(QueueItemActionImpl.class);
```

jglick: never mind, forgot about https://github.com/jenkinsci/workflow-api-plugin/blob/89f199273075741ba6d3728a639126de7914febd/src/main/java/org/jenkinsci/plugins/workflow/actions/QueueItemAction.java#L13

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecutionRJRTest.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Rule public RealJenkinsRule rjr = new RealJenkinsRule().withColor(PrefixedOutputStream.Color.GREEN).withPackageLogger(ExecutorStepExecution.class, Level.FINE).withPackageLogger(FlowExecutionList.class, Level.FINE);
+
+    @Rule
+    public InboundAgentRule iar = new InboundAgentRule();
+
+    @Test public void restartWhileWaitingForANode() throws Throwable {
+        rjr.startJenkins();
+        try (var tailLog = new TailLog(rjr, "p", 1)) {
+            iar.createAgent(rjr, InboundAgentRule.Options.newBuilder().name("J").label("mib").color(PrefixedOutputStream.Color.YELLOW).webSocket().build());
```
jglick, line 69: BTW do _J_ and _mib_ mean something?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecutionRJRTest.java
```diff
@@ -0,0 +1,149 @@
...
+                        input id:'Branch2', message: 'Complete branch 2 ?'
+                    }
+                }
+                """, true));
+        var b = p.scheduleBuild2(0).waitForStart();
+        r.waitForMessage("Complete branch 1 ?", b);
+        assertTrue(b.isBuilding());
+        await().until(() -> r.jenkins.getQueue().getItems(), arrayWithSize(1));
+        LOGGER.info("Node steps: " + new DepthFirstScanner().filteredNodes(b.getExecution(), new NodeStepTypePredicate("node")));
+        // "Branch 1", "Branch 2" and the second node step ? Not fully clear
```
jglick, line 146:
```suggestion
        // "Branch 1" step start + "Branch 1" body start + "Branch 2" step start
```
?

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecutionRJRTest.java
```diff
@@ -0,0 +1,149 @@
...
+            rjr.stopJenkinsForcibly();
+            rjr.startJenkins();
+            rjr.runRemotely(ExecutorStepExecutionRJRTest::resumeCompleteBranch1ThenBranch2);
+        }
+    }
+
+    private static void resumeCompleteBranch1ThenBranch2(JenkinsRule r) throws Throwable {
+        var p = r.jenkins.getItemByFullName("p", WorkflowJob.class);
+        var b = p.getBuildByNumber(1);
+        await("Waiting for agent J to reconnect").atMost(Duration.ofSeconds(30)).until(() -> r.jenkins.getComputer("J").isOnline(), is(true));
```
jglick, line 80: or just
```suggestion
        await("Waiting for agent J to reconnect").atMost(Duration.ofSeconds(30)).until(() -> r.jenkins.getComputer("J").isOnline());
```
https://javadoc.io/static/org.awaitility/awaitility/4.2.2/org/awaitility/core/ConditionFactory.html#until-java.util.concurrent.Callable-

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecutionRJRTest.java
Vlatombe: Reference to Men In Black, special Agent J

src/test/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecutionRJRTest.java
```diff
@@ -0,0 +1,149 @@
...
+        var actions = await().until(() -> b.getActions(InputAction.class), allOf(iterableWithSize(1), hasItem(new InputActionWithId("Branch1"))));
+        proceed(actions, "Branch1", p.getName() + "#" + b.number);
+        // This is quicker than waitForMessage that can wait for up to 10 minutes
```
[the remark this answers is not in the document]
Vlatombe: Yes, would be much nicer than the current assertions available on `JenkinsRule`.

jglick (approved):

PR #463, opened 2025-07-22
Refining automatic aborts of `DurableTaskStep`

jglick:


src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -692,6 +696,20 @@ private static final class HelloNote extends ConsoleNote<Run<?, ?>> {
...
+        DurableTaskStep.WATCHING_RECURRENCE_PERIOD = Duration.ofSeconds(30).toMillis();
+        logging.record(DurableTaskStep.class, Level.FINE);
+        var s = j.createSlave("remote", null, null);
+        s.setRetentionStrategy(RetentionStrategy.NOOP);
+        var p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node('remote') {isUnix() ? sh('sleep 1000000') : bat('ping -t 127.0.0.1 > nul')}", true));
+        var b = p.scheduleBuild2(0).waitForStart();
+        j.waitForMessage(Functions.isWindows() ? ">ping" : "+ sleep", b);
+        s.toComputer().disconnect(null);
+        j.assertBuildStatus(Result.ABORTED, j.waitForCompletion(b));
```
jglick, line 709: Times out without fix.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -692,6 +696,36 @@ private static final class HelloNote extends ConsoleNote<Run<?, ?>> {
...
+        j.waitForMessage(new ExecutorStepExecution.RemovedNodeTimeoutCause().getShortDescription(), b);
+    }
+
+    @Test public void temporarilyDisconnectingAgentIsNotFatal() throws Exception {
+        DurableTaskStep.WATCHING_RECURRENCE_PERIOD = Duration.ofSeconds(30).toMillis();
+        logging.record(DurableTaskStep.class, Level.FINE);
+        var s = j.createSlave("remote", null, null);
+        s.setRetentionStrategy(RetentionStrategy.NOOP);
+        var p = j.jenkins.createProject(WorkflowJob.class, "p");
+        p.setDefinition(new CpsFlowDefinition("node('remote') {isUnix() ? sh('sleep 60') : bat('ping -c 60 -t 127.0.0.1 > nul')}", true));
```
jtnord, line 719: DO you want to ping 60 times?  
`-c` is for routing and requires admin access.

```
❯ where ping
C:\Windows\System32\PING.EXE

  D:\s\github  ☸ docker-desktop default  where ping^C                                                   13:51:00 
❯ ping -c 60 -t 127.0.0.1
Access denied. Option -c requires administrative privileges.

❯ ping --help
Bad option --help.

Usage: ping [-t] [-a] [-n count] [-l size] [-f] [-i TTL] [-v TOS]
            [-r count] [-s count] [[-j host-list] | [-k host-list]]
            [-w timeout] [-R] [-S srcaddr] [-c compartment] [-p]
            [-4] [-6] target_name

Options:
    -t             Ping the specified host until stopped.
                   To see statistics and continue - type Control-Break;
                   To stop - type Control-C.
    -a             Resolve addresses to hostnames.
    -n count       Number of echo requests to send.
    -l size        Send buffer size.
    -f             Set Don't Fragment flag in packet (IPv4-only).
    -i TTL         Time To Live.
    -v TOS         Type Of Service (IPv4-only. This setting has been deprecated
                   and has no effect on the type of service field in the IP
                   Header).
    -r count       Record route for count hops (IPv4-only).
    -s count       Timestamp for count hops (IPv4-only).
    -j host-list   Loose source route along host-list (IPv4-only).
    -k host-list   Strict source route along host-list (IPv4-only).
    -w timeout     Timeout in milliseconds to wait for each reply.
    -R             Use routing header to test reverse route also (IPv6-only).
                   Per RFC 5095 the use of this routing header has been
                   deprecated. Some systems may drop echo requests if
                   this header is used.
    -S srcaddr     Source address to use.
    -c compartment Routing compartment identifier.
    -p             Ping a Hyper-V Network Virtualization provider address.
    -4             Force using IPv4.
    -6             Force using IPv6.


```


```suggestion
        p.setDefinition(new CpsFlowDefinition("node('remote') {isUnix() ? sh('sleep 60') : bat('ping -r 60 -t 127.0.0.1 > nul')}", true));
```

jtnord (changes requested):

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
jtnord: TIL `timeout.exe`  https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/timeout  should be available in Windows Server also.

(untested other than the command runs locally)

```suggestion
        p.setDefinition(new CpsFlowDefinition("node('remote') {isUnix() ? sh('sleep 60') : bat('timeout /T 60 /NOBREAK')}", true));
```

jglick:
> I can see cases where you have a long running job and there is some network maintenance. […] At the end of the day the node is still defined so we would expect it to come back (for static agents. or be cleaned up by a cloud for dynamic agents) wouldn't we?

Maybe? On the other hand the _current_ behavior is that the build gets aborted at the next controller restart.

If you were doing network maintenance, would you not likely also be maintaining the machines and perhaps rebooting them? You can always `retry` with `conditions` if this is an occasionally expected scenario.

The root problem in at least some cases (permanent SSH agent, EC2 cloud agent) does indeed seem to have been an incorrectly implemented `RetentionStrategy` that failed to `check` whether the agent was offline and relaunch. But I also saw similar behavior under load in a K8s agent; maybe the pod kept running but never managed to reconnect for whatever reason. Ideally the agent would get deleted by cloud-specific code in such circumstances but in case it does not, you are stuck with a hanging build.

dwnusbaum (approved):

jglick:
> Previously, the jobs were finishing as `FAILURE`, so we caught that and retried them. Now _some of them_ are finishing as `ABORTED`

Well this PR would make the `sh` fail with `ABORTED`. Not sure what would have caused `FAILURE` before specifically (would depend on details).

> we don't want to retry jobs that were aborted by a person, or by a calling pipeline, etc.

Right, so use

```groovy
retry(count: 2, conditions: [nonresumable(), agent()]) {
  node(…) {
    sh '…'
  }
}
```

PR #464, opened 2025-07-28
Inherit describable logstorage from workflow-api

sandeepbarnwal:
> Note: It is still in draft. Please don't review this for now. 

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


jgreffe:
https://github.com/jenkinsci/workflow-api-plugin/pull/417

pom.xml
```diff
@@ -95,6 +95,8 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins.workflow</groupId>
             <artifactId>workflow-api</artifactId>
+            <!-- TODO: Replace incremental version with latest release once upstream is released -->
+            <version>1397.v02e97a_30f3ce</version>
```
dwnusbaum, line 99: @sandeepbarnwal when you need to update a plugin like this whose version normally comes from the BOM, use `dependencyManagement` instead. While waiting for the upstream PR to be merged, I like to add a comment that mentions the other PR, to make it clear where this comes from, e.g. `TODO: https://github.com/jenkinsci/workflow-api-plugin/pull/417`. I would also mention that PR in the PR description here (Julien already mentioned it in a comment, but this is a good thing to get in the habit of doing.)

Once the upstream PR is merged, you can remove that TODO comment, but instead replace it with one that says something like `TODO: Remove once this version is included in the BOM`, so next time someone updates the BOM, they can check to see if they can delete the `dependencyManagement` entry completely.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -458,7 +465,22 @@ private static class Decorator extends LauncherDecorator implements Serializable
                 }
             };
         }
+
+        @TestExtension("remoteLogger[watching=true]")
+        public static final class DescriptorImpl extends LogStorageFactoryDescriptor<LogFile> {
+            @NonNull
+            @Override
+            public String getDisplayName() {
```
Amankkc, line 473: As this is in the test scope , we can remove getDisplayName IMO

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -431,8 +432,14 @@ private static class Decorator extends LauncherDecorator implements Serializable
         j.assertLogNotContains(password.toUpperCase(Locale.ENGLISH), b);
         j.assertLogContains("CURL -U **** HTTP://SERVER/ [master → remote]", b);
     }
-    @TestExtension("remoteLogger[watching=true]") public static class LogFile implements LogStorageFactory {
-        @Override public LogStorage forBuild(FlowExecutionOwner b) {
+
+    public static class LogFile implements LogStorageFactory {
+
+        @DataBoundConstructor
+        public LogFile() {}
```
Amankkc, line 439: use 

```
public LogFile() { 
  //noop
}
```

This will be raised by sonarqube as an issue but as OSS dont have sonarqube integrated so it was not raised . see this -> https://github.com/cloudbees/cloudbees-replication/pull/756/commits/10236147ab0084435c2c83d391690f2566b079f2

Amankkc: Even this DataBoundConstructor can be removed as it is getting used in the test scope IMO

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
sandeepbarnwal: I actually included that intentionally to give it more meaning.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
sandeepbarnwal: Since there's no SonarQube involved, I suppose it's fine as it is.

src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
Amankkc: Adding this suggests that the name will be displayed in the UI, but here, it's used within the test scope. IMO, it would be better to remove the redundant code, and if you want to add more context, using comments might be a better approach.

sandeepbarnwal: That’s not quite how I see it. JenkinsRule does spin up an actual Jenkins instance, so I’m not sure why it'd be wrong that name is displayed in the UI.
Also, if the code itself can make the intent clear, then a comment might not be necessary or ideal.
That said, I don’t have a strong opinion on this and I’m fine either way.
@jgreffe @dwnusbaum wdyt?

dwnusbaum: Ultimately I think it doesn't really matter either way. Unless you access the Jenkins instance in your browser while running tests, you'll probably never notice the display name. Customizing it does not add any real complexity, but the default value of `LogStorage` would be just as clear as the customized value, so there's not much point to customizing it. I would lean towards just deleting the override in test-specific implementations, unless the tests actually use the display names.

PR #472, opened 2025-08-18
`ContinuedTask` cannot rely on queue scheduling order

jglick:
There seems to be a race condition with `ContinuedTask`: its effectiveness relies on the `PlaceholderTask` that was continued having actually been scheduled (and not merely in `pendings` but in `buildables`) by the time other tasks are added to the queue. That is not a safe assumption at least since https://github.com/jenkinsci/workflow-api-plugin/pull/221 (and possibly even before that), and certainly as of https://github.com/jenkinsci/workflow-api-plugin/pull/368 this is not a good basis for blocking queue items. I think it is necessary to somehow mark the `Slave` prior to shutdown as having been used for a given build, and wait for it to be ready before considering scheduling anything else onto the agent. ~(But not using anything in `config.xml`, since that would be clobbered by JCasC in the case of permanent agents.) Perhaps there is some simpler criterion that can be used.~

[CloudBees-internal reference](https://cloudbees.atlassian.net/browse/BEE-61733)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepDynamicContext.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.workflow.support.steps;
 
+import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
 import hudson.Extension;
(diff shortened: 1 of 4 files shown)


dwnusbaum:
Maybe a `QueueTaskDispatcher` that stores persistent state about agents being used by `node` steps could help, since that would be able to take effect prior to the step resuming?

jglick:
Related: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/382

jglick:
This approach seems to work, though I do not love the idea of writing `config.xml` for a permanent agent every time a `node` block starts or stops. At this point the bug in OSS remains theoretical (only observed under heavy load in a CloudBees CI HA controller). I may park this.

PR #474, opened 2025-09-09
Migrate Commons Lang from 2 to 3 and StringEscapeUtils to Commons Text

gounthar:
Hello `workflow-durable-task-step` developers! :wave:

This is an automated pull request created by the [Jenkins Plugin Modernizer](https://github.com/jenkins-infra/plugin-modernizer-tool) tool. The tool has applied the following recipes to modernize the plugin:
<details aria-label="Recipe details for Migrate Commons Lang from 2 to 3 and StringEscapeUtils to Commons Text">
    <summary>Migrate Commons Lang from 2 to 3 and StringEscapeUtils to Commons Text</summary>
    <p><em>io.jenkins.tools.pluginmodernizer.MigrateCommonsLang2ToLang3AndCommonText</em></p>
    <blockquote>Migrate Commons Lang from 2 to 3 and StringEscapeUtils to Commons Text.</blockquote>
</details>

This pull request upgrades `Apache Commons Lang 2` to `Apache Commons Lang 3` and migrates HTML escaping functionality from `Apache Commons Lang` to `Apache Commons Text`.
The `Apache Commons` community recommends using `Apache Commons Text` for string escaping operations, as it offers a more focused and feature-rich API for text processing.

## What's Changed?

- Migrated from deprecated/EOL `Apache Commons Lang 2` to `Commons Lang 3`.
- Transitioned from deprecated/EOL `StringEscapeUtils` in `Apache Commons Lang` to `Commons Text`.

PR #475, opened 2025-09-09
Avoiding `StringUtils`

jglick:
https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/474#pullrequestreview-3202538591

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -69,7 +69,6 @@ import jenkins.model.Jenkins;
 import jenkins.tasks.filters.EnvVarsFilterableBuilder;
 import jenkins.util.Timer;
 import org.apache.commons.io.IOUtils;
-import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.durabletask.Controller;
 import org.jenkinsci.plugins.durabletask.DurableTask;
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -149,8 +148,8 @@ public String getLabel() {
     }
 
     @Override public StepExecution start(StepContext context) throws Exception {
-        if (this.label != null) {
-            context.get(FlowNode.class).addAction(new LabelAction(StringUtils.left(label, MAX_LABEL_LENGTH)));
+        if (label != null) {
+            context.get(FlowNode.class).addAction(new LabelAction(label.length() > MAX_LABEL_LENGTH ? label.substring(0, MAX_LABEL_LENGTH) : label));
```
jglick, line 152: Dubious logic from #93; IMO better for any truncation to be done in display plugins, not here.

src/test/java/org/jenkinsci/plugins/workflow/EnvWorkflowTest.java
```diff
@@ -53,21 +52,23 @@ public class EnvWorkflowTest {
      */
     @Test public void isNodeNameAvailable() throws Exception {
         r.createSlave("node-test", "unix fast", null);
-        String builtInNodeLabel = r.jenkins.getSelfLabel().getExpression(); // compatibility with 2.307+
```
jglick, line 56: Old.

PR #478, opened 2025-09-16
Optimized `AgentReconnectionListener`

jglick:
Listener introduced in c69c1aa7d6fb99a54b98137c65d509e13b9f25cd (#323) was too expensive when a lot of builds are running at once.


src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -782,12 +789,33 @@ private static class HandlerImpl extends Handler {
...
+                    var feo = feoe.asFlowExecutionOwner();
+                    if (feo != null) {
+                        try {
+                            var fe = feo.get();
+                            var executionsFuture = fe.getCurrentExecutions(true);
+                            executionsFuture.addListener(() -> {
+                                try {
+                                    for (var execution : executionsFuture.get()) {
+                                        if (execution instanceof Execution exec && exec.watching && exec.node.equals(name)) {
+                                            LOGGER.fine(() -> "Online/offline event on " + name + ", checking current status of " + exec.remote + " soon");
```
jglick, line 805: (ignore WS)

PR #620, opened 2026-05-11
Honor `causeOfStoppage` even when process exits with code 0

### LINKED ISSUE #619: `causeOfStoppage` ignored by `handleExit` if `exitCode == 0`
#75 (amended by #84) tries to ensure that if `stop` is called on a `sh` step, `StepContext.onFailure` will in fact reflect that cause, producing an appropriate build status and `CauseOfInterruption` message. But there is a race condition here: if the shell script happened to complete normally before the `SIGTERM` could even be delivered, the step will succeed and the `cause` will be discarded.

I was able to reproduce this in a test which called `FlowExecution.interrupt` in response to certain log text being printed by a very short-lived shell script (adding a delay of just a second solves it). Besides manual abort, I suspect that this could happen occasionally with other Pipeline features. For example, if a `timeout` wraps `sh` and the timeout happens to expire within some milliseconds of when the script completed (perhaps even on a scale of seconds, if `!USE_WATCHING` and `recurrencePeriod` has grown due to inactivity), the corresponding `ExceededTimeout` might be lost. Thus _Cancelling nested steps due to timeout_ would be printed but the serious problem is that the CPS program is not interrupted: other steps will run normally! 1m later _Body did not finish within grace period; terminating with extreme prejudice_ will be printed and the build will probably abort then, though not gracefully.

Needs testing but I think the fix in https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/ed562f6ec8b3eafb64cf27d5326a6adf4beca3ed/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L681-L691 would be to ensure that if `originalCause != null` then we always `getContext().onFailure(originalCause)`. Would need to study the expected interactions with `returnStatus` and `returnStdout` options.

Possibly related: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/84#discussion_r227814515

jglick:
Fixes #619. Amending #75.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
+++ b/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
@@ -678,20 +678,19 @@ public abstract class DurableTaskStep extends Step implements EnvVarsFilterableB
             byte[] produce() throws IOException, InterruptedException;
         }
         private void handleExit(int exitCode, OutputSupplier output) throws IOException, InterruptedException {
-            Throwable originalCause = causeOfStoppage;
-            if ((returnStatus && originalCause == null) || exitCode == 0) {
-                getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(output.produce(), StandardCharsets.UTF_8) : null);
-            } else {
+            if (causeOfStoppage != null) {
+                getContext().onFailure(causeOfStoppage);
+            } else if (returnStatus) {
+                getContext().onSuccess(exitCode);
+            } else if (exitCode != 0) {
                 if (returnStdout) {
                     _listener().getLogger().write(output.produce()); // diagnostic
                 }
-                if (originalCause != null) {
-                    // JENKINS-28822: Use the previous cause instead of throwing a new AbortException
-                    _listener().getLogger().println("script returned exit code " + exitCode);
-                    getContext().onFailure(originalCause);
-                } else {
-                    getContext().onFailure(new AbortException("script returned exit code " + exitCode));
-                }
+                getContext().onFailure(new AbortException("script returned exit code " + exitCode));
+            } else if (returnStdout) {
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/workflow/steps/durable_task/ShellStepTest.java
```diff
@@ -778,7 +778,7 @@ private static final class HelloNote extends ConsoleNote<Run<?, ?>> {
         j.waitForCompletion(b);
         // Would have succeeded before https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/75.
         j.assertBuildStatus(Result.ABORTED, b);
-        j.waitForMessage("Timeout has been exceeded", b); // TODO assertLogContains fails unless a sleep is introduced; possible race condition in waitForCompletion
+        j.assertLogContains("Timeout has been exceeded", b);
```
jglick, line 781: reverting https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/84#discussion_r227814515

src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java
```diff
@@ -678,20 +678,19 @@ private interface OutputSupplier {
             byte[] produce() throws IOException, InterruptedException;
         }
         private void handleExit(int exitCode, OutputSupplier output) throws IOException, InterruptedException {
-            Throwable originalCause = causeOfStoppage;
-            if ((returnStatus && originalCause == null) || exitCode == 0) {
-                getContext().onSuccess(returnStatus ? exitCode : returnStdout ? new String(output.produce(), StandardCharsets.UTF_8) : null);
-            } else {
+            if (causeOfStoppage != null) {
+                getContext().onFailure(causeOfStoppage);
+            } else if (returnStatus) {
```
jglick, line 683: Trying to make the logic of this method more legible.

jtnord (approved):

PR #624, opened 2026-06-06
[JENKINS-60507] Do not strand runs / pin executors on transient "did not yet start" during restart

VetleNeumann:
## Problem

See JENKINS-60507 for the full diagnosis. Summary: during a controller restart window, `WorkflowRun$Owner.get` throws `IOException: "<run> did not yet start"` while the run is still loading. Two code paths in `ExecutorStepExecution` mishandle that transient exception and permanently wedge the build:

1. `PlaceholderTask.withExecution(...)` caught and **swallowed** the `IOException` from `context.get(FlowExecution.class)`. The async callback never fired, the body never started, and the caller registered an `AsynchronousExecution` that **pins the executor forever**.
2. The deprecated `PlaceholderTask.run()` accessor let the same transient `IOException` fall into a generic `catch` that called `finish()` + `RunningTasks.remove()`, **tearing down the placeholder** while the CPS `node` step was still live — **stranding the run IN_PROGRESS** with the branch stuck `QUEUED`. This is the "node block appears to be neither running nor scheduled" symptom from the original report.

A third, contention-related gap: `withExecution`'s async lookup could resolve before the matching `ExecutorStepExecution` was registered and then silently do nothing.

## Changes

- `withExecution` now declares `throws IOException, InterruptedException` instead of swallowing — the caller is responsible for failing the step and freeing the executor. Call sites updated to handle/log.
- The async lookup is split into `withExecutionAsync`, which retries (up to `WITH_EXECUTION_MAX_ATTEMPTS`, 500 ms apart via `Timer`) when the result set doesn't yet contain the matching execution, and logs at `SEVERE` if it exhausts retries.
- `PlaceholderTask.run()` now catches `IOException` separately, logs at `FINE`, and returns `null` **without** tearing the task down — the not-yet-loaded condition is transient. Genuine breakage still falls through to the existing teardown path.

## Tests

A straightforward "restart with parallel `node` blocks" test does not reliably hit the timing window and passes on both unpatched and patched builds, so it doesn't lock down the fix. These three tests instead drive each failure path deterministically via fault injection (a `StepContext` armed to throw the transient `IOException`), and are all discriminating — **red on `master`, green with this change** (both verified):

- `runAccessorDoesNotTearDownTaskWhenRunNotYetLoaded` — the placeholder survives a transient `"did not yet start"` (pre-fix it was removed, stranding the run).
- `withExecutionPropagatesTransientFlowExecutionLoadFailure` — the load failure reaches the caller (pre-fix it was swallowed, pinning the executor).
- `withExecutionRetriesUntilMatchingExecutionAppears` — the async lookup retries until the matching execution registers under contention (pre-fix it gave up silently on the first miss).

Not covered: the `cancelOwnerExecution` / interrupt call sites — same try/catch shape, low marginal value.

## Likely related reports

These appear to share this root cause and may be duplicates: #567 (JENKINS-54513, "Builds occupying executor but not running"), #598 (JENKINS-72087, "Hangs of executors" — `getCurrentExecutable().getParent().run()` returns `null` from the script console, exactly the torn-down accessor), #504 (JENKINS-39552, "After restart, interrupted pipeline deadlocks waiting for executor"). #582 (JENKINS-64412) is an adjacent startup-race symptom.


VetleNeumann:
For anyone curious how we ended up recreating this: the normal way you'd expect to repro it - parallel `node` blocks, restart mid-build, watch a branch hang - just never tripped it for us on stock binaries. We tried a ton of variations (fresh containers, killing queue saturation, hammering the `Owner.get` window, agent disconnects mid-run, rapid restart loops) and got nothing, every time.

The catch is that a clean restart always reloads the run record before it resumes placeholders, so `WorkflowRun$Owner.get` never actually throws the `"<run> did not yet start"` that sets the whole thing off. Restart alone isn't enough - you also need the record to still be unloadable at the moment a placeholder gets interrupted, which in the one real capture we had came from a shutdown interrupt rather than the resume path. Once that clicked we stopped chasing the timing and just injected that exception where it'd really fire, which is what the tests do.

To be upfront about scope: this doesn't explain *why* a run record ends up unloadable in prod in the first place - that part's still open. What it does is make sure that when it does happen, the placeholder fails cleanly and frees the executor instead of stranding the build forever. Handling the broken state gracefully, rather than pretending we've mapped every road into it.

